PR #21, opened 2026-05-06
feat: add automatic release changelog with release-please

### LINKED ISSUE #17: Add automatic release changelog generator
Need to look into some options, but something like changesets or semantic release would be nice. Ideally, something that plays well with goreleaser, as well.

We should link to the changelog in our update prompts / script as well.

blimmer:
Closes #17.

## Summary

Adds release-please as the source of truth for stable-release versioning and `CHANGELOG.md`, while keeping the existing goreleaser pipeline as the artifact-distribution layer. release-please opens a Release PR on each push to `main`; merging it creates the tag and a GitHub Release whose body is the new CHANGELOG entry. Goreleaser then runs on the tag push with `release.mode: keep-existing`, attaching tarballs and casks without disturbing the body. Manual alpha tags continue to bypass release-please entirely. The CLI's update prompts and the plan UI's update card now link to the per-version GitHub Release page so users can see what changed before upgrading.

## Review focus

- **release-please / goreleaser composition.** Both tools can create a GitHub Release; the coordination knob is goreleaser's `release.mode: keep-existing` (`.goreleaser.yaml`). Without it, goreleaser would clobber release-please's curated body. Worth eyeballing the order of operations described in `release-please-config.json` + `release-please.yml` against the existing tag-triggered `release.yml`.
- **`release-type: go` for a TypeScript repo.** The label is misleading but the behavior fits: it bumps `CHANGELOG.md` only, not any version file. We don't want package.json drifting from `0.0.0-development` since goreleaser injects the real version from the tag.
- **Manifest seeded at `0.2.0`.** Pinning matches the most recent stable tag so the first release-please run computes from `v0.2.0` forward, not from repo inception.
- **Action SHA pinning.** All three new actions (`amannn/action-semantic-pull-request`, `googleapis/release-please-action`, `actions/create-github-app-token`) are SHA-pinned with the version as a trailing comment. Matches existing pattern in `release.yml`.
- **GH App token in `release-please.yml`.** Same `CB_PR_AUTOMATION_APP_*` pair already used by `release.yml`. Required so the tag release-please pushes triggers downstream workflows — a default `GITHUB_TOKEN` would not.
- **`UpdateNoticeCard` link placement.** First pass put a "What's new" link mid-description, which wrapped awkwardly in the 20rem card. Final design makes `v{version}` in the title itself the link — same compact 3-line layout as before.

## Commits

- [`5e29026`](https://github.com/contextbridge/planbridge/commit/5e29026) — feat: add automatic release changelog with release-please
- [`5ee1908`](https://github.com/contextbridge/planbridge/commit/5ee1908) — refactor: use shared `GITHUB_REPO_URL` for release links
- [`3d07f37`](https://github.com/contextbridge/planbridge/commit/3d07f37) — docs: document release process in CONTRIBUTING


.github/workflows/lint-pr-title.yml
```diff
@@ -0,0 +1,37 @@
+name: Lint PR title
```
blimmer, line 1: I will add this to the required checks once merged. It ensures that people follow conventional-commit so our changelog works.

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,33 @@
...
+jobs:
+  release-please:
+    name: release-please
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - name: Generate ContextBridge app token
```
blimmer, line 19: Needs to be able to write to the repo, so using our app

release-please-config.json
```diff
@@ -0,0 +1,25 @@
...
+        { "type": "perf", "section": "Performance" },
+        { "type": "deps", "section": "Dependencies" },
+        { "type": "revert", "section": "Reverts" },
+        { "type": "docs", "section": "Documentation", "hidden": true },
+        { "type": "chore", "section": "Miscellaneous", "hidden": true },
+        { "type": "style", "section": "Styles", "hidden": true },
+        { "type": "refactor", "section": "Refactors", "hidden": true },
+        { "type": "test", "section": "Tests", "hidden": true },
+        { "type": "build", "section": "Build", "hidden": true },
+        { "type": "ci", "section": "CI", "hidden": true }
```
blimmer, line 21: Thoughts on hidden sections? Does this seem like a good starting place?

jcarver989: Yea think this looks like a good starting point in that we want user-facing changes visible but refactoring etc not.

jcarver989 (approved):
Looks reasonable, thanks!

PR #48, opened 2026-05-07
fix: emit plan review analytics from shared runner

### LINKED ISSUE #47: bug: plan_review_submitted is not emitted for hook-triggered reviews
## What happened?

Plan reviews submitted through harness hooks did not emit the `plan_review_submitted` analytics event. The browser UI still emitted `plan_review_viewed`, so reviews appeared to be opened but never submitted in PostHog. This undercounted daily active users for users whose normal workflow is Claude Code or Codex hook-triggered review.

## contextbridge version

0.2.0

## Operating system

macOS

## Steps to reproduce

1. Install or configure PlanBridge for Claude Code or Codex hooks.
2. Trigger a plan review via the hook path, not by running `contextbridge plan` directly.
3. Submit or approve the plan review in the browser UI.
4. Inspect PostHog for the submitting user events.

## What did you expect to happen?

Submitting a plan review should emit `plan_review_submitted` regardless of whether the review was triggered by `contextbridge plan`, `contextbridge hook claude`, `contextbridge hook codex`, or a future hook adapter.

## What actually happened?

For hook-triggered reviews, `plan_review_viewed` continued to appear from the browser UI, but `plan_review_submitted` was missing. The direct `contextbridge plan` path emitted `plan_review_submitted`, because the analytics capture lived in the direct command adapter instead of the shared plan-review runner used by hooks.

## Relevant logs or stack trace

No CLI error output. The failure was visible in PostHog event data: recent hook-triggered sessions had browser-side `plan_review_viewed` events, but no corresponding `plan_review_submitted` events.

## Browser and version

Not browser-specific.

## Anything else

Root cause: plan lifecycle analytics were tied to the direct `runPlan` command path. Hook adapters call `runPlanReview` directly, so lifecycle events need to live at the shared plan-review boundary.

blimmer:
## Summary

Plan review lifecycle analytics now live in the shared `runPlanReview` path instead of only the direct `contextbridge plan` command. Hook-triggered Claude and Codex reviews pass explicit sources into that runner, so successful submissions emit `plan_review_submitted` regardless of how the review was launched.

Fixes #47.

## Review focus

Please check the source values (`stdin`, `file`, `hook_claude`, `hook_codex`) and the placement of `plan_review_started` / `plan_review_submitted` in `runPlanReview`. The submit event is still skipped for open failures and SIGINT interruptions.

## Commits

- [`3ea2cf3`](https://github.com/contextbridge/planbridge/commit/3ea2cf33f8815a474b1b5853766ac380ca095a53) — moves plan review lifecycle analytics into the shared runner and adds direct/hook coverage.


packages/cli/src/commands/plan.ts
```diff
@@ -40,16 +38,10 @@ export async function runPlan(ctx: CliContext, args: PlanArgs, deps?: PlanReview
   }
 
   logger.info({ source, bytes: Buffer.byteLength(content, 'utf8') }, 'plan received');
-  analytics.capture('plan_review_started', { source });
 
   try {
-    const submission = await runPlanReview(ctx, { planContent: content }, deps);
+    const submission = await runPlanReview(ctx, { planContent: content, source }, deps);
     io.stdout.write(formatAsMarkdown(submission, content));
-    analytics.capture('plan_review_submitted', {
```
blimmer, line 48: This was only firing when directly triggering via `contextbridge plan` before, not the hook.

packages/cli/src/planReview/runPlanReview.ts
```diff
@@ -80,7 +84,13 @@ export async function runPlanReview(
     logger.info({ url: server.url }, 'opening plan-review browser session');
     await Promise.race([openUrl(server.url), sigintPromise]);
 
-    return await Promise.race([server.result, sigintPromise]);
+    const submittedReview = await Promise.race([server.result, sigintPromise]);
+    analytics.capture('plan_review_submitted', {
```
blimmer, line 88: sigint rejects the promise, so we won't send this if sigint occurs.

jcarver989 (approved):

Issue #51, opened 2026-05-07
feat: skill to open an arbitrary file in PlanBridge


blimmer:
### What's the problem?

PlanBridge currently only opens when a harness hook fires on `ExitPlanMode` (Claude Code) or the equivalent Codex hook. There is no way for a user to manually open an arbitrary file — a draft plan, an architecture doc, a checklist — in the PlanBridge UI for annotation.

Two concrete user pains today:

* Users who keep design docs or in-flight plans as Markdown in their repo cannot pull them into PlanBridge for review without re-driving the agent through plan mode.
* The [superpowers](<https://github.com/obra/superpowers>) plugin's skill flow does not call `ExitPlanMode` and instead writes plan files directly to disk, so PlanBridge hooks never fire. See contextbridge/planbridge#37 and discussion [#50](<https://github.com/contextbridge/planbridge/issues/50>). A manual command would let those users trigger annotation on the file superpowers just wrote.

### Proposed solution

Add a Claude Code slash command, shipped in the plugin manifest being moved in-tree by contextbridge/planbridge#49, that takes a file path argument and opens that file in the PlanBridge browser UI.

Layered as:

1. **New CLI subcommand** — `contextbridge open <path>` (or whatever verb wins, see Alternatives) — that loads the file, starts the local server, and opens the UI. Mirrors the existing `contextbridge plan` shape so non-Claude harnesses get the same capability.
2. **Slash command in the plugin** — `/planbridge-open <path>`, a thin wrapper that passes its argument to the CLI subcommand. Lives alongside the hook definitions being relocated by contextbridge/planbridge#49. It should support human language description, too, e.g., the name of the plan in the canonical superpowers location (the LLM should decide here the file is - the `open` command should be a place on disk).

### Alternatives you considered

**Naming.** `/planbridge-review` was the obvious first try, but "review" is the verb we want to reserve for the planned diff/code-review feature where it carries the conventional "review of proposed changes" meaning (see [https://github.com/contextbridge/planbridge/discussions/29](<https://github.com/contextbridge/planbridge/discussions/29>)). Other candidates considered:

Recommended:

* `/planbridge-open <path>` — neutral and action-agnostic; "open the file in PlanBridge" leaves room for whatever the user wants to do next (comment, approve, request changes); doesn't lock the verb to a single interaction mode.

Other candidates:

* `/planbridge-comment <path>` — matches the verb in the original feature request; understated since PlanBridge also gates approval.
[...]

PR #52, opened 2026-05-07
feat!: rename Claude plugin to planbridge@contextbridge; refresh plugins on update

### LINKED ISSUE #49: feat: move Claude Code plugin assets into this repo, rename to planbridge@contextbridge
### What's the problem?

PlanBridge's Claude Code integration is split across two repos:

- **`contextbridge/planbridge`** (this repo) — the `contextbridge` binary, including the `contextbridge hook claude` subcommand the plugin actually invokes, plus the `ClaudeInstaller` that wires Claude Code up via `claude plugin marketplace add` / `claude plugin install`.
- **`contextbridge/claude-plugin`** — `marketplace.json` (the catalog), `plugin.json`, and `hooks/hooks.json` declaring `PermissionRequest:ExitPlanMode → contextbridge hook claude`.

The original reason for the split was that this repo wasn't public — the plugin manifest had to live somewhere users could `claude plugin marketplace add`, so it lived in a separate public repo. That constraint is gone, but two specific files (`plugin.json`, `hooks/hooks.json`) being in the wrong repo still costs us:

- **Drift risk.** `hooks.json`'s command string (`contextbridge hook claude`) lives in one repo and the `hookClaude.ts` implementation lives in the other. Nothing in CI cross-validates them; a rename in one repo silently breaks the other.
- **Two-repo release dance.** Any change touching both the hook contract and the manifest needs coordinated PRs across repos.
- **Discoverability.** A contributor reading this repo can't see how Claude Code actually invokes the CLI without knowing the other repo exists.

While we're in here: the plugin is named `cli`, so its install id is `cli@contextbridge`. That made sense when `cli` was the only thing — it doesn't anymore. The product is PlanBridge, not "the CLI," and once sibling plugins ship (code review, etc.) `cli` becomes meaningless as a distinguishing name.

There's also a deeper problem the move surfaces: today `contextbridge update` swaps the binary and stops there. That works only because the hook contract is `command: "contextbridge hook claude"` — a PATH lookup that picks up the new binary automatically. The moment the plugin id, `hooks.json`, or `plugin.json` `version` changes (this PR or future hook-contract changes), the binary and plugin drift. Worth fixing as part of this PR: couple binary and plugin updates.

### Proposed solution

Two changes, landing together. Take the breaking change on the plugin rename now while the user base is small — much cheaper than doing it later.

**Keep `contextbridge/claude-plugin` as the marketplace catalog.** It's the storefront where users discover all contextbridge plugins (today: PlanBridge; future: code review, etc.). Each plugin's actual source lives in its own product repo. Claude Code's marketplace schema supports cross-repo plugin sources, so this is straightforward.

**Move plugin assets into this repo:**

```
harnessIntegrations/claude/.claude-plugin/plugin.json
harnessIntegrations/claude/hooks/hooks.json
```

(Top-level folder is `harnessIntegrations` so future Codex / other harness assets land as siblings — `harnessIntegrations/claude/`, future `harnessIntegrations/codex/`, etc.)

**Rename the plugin** in the new `plugin.json`: `"name": "cli"` → `"name": "planbridge"`. New install id is `planbridge@contextbridge`.

**Edit `contextbridge/claude-plugin/.claude-plugin/marketplace.json`** to point at this repo via a cross-repo source:

```diff
   "plugins": [
     {
-      "name": "cli",
-      "source": "./cli",
-      "description": "ContextBridge Claude Code integration — opens plan reviews in the browser on ExitPlanMode."
+      "name": "planbridge",
+      "source": { "source": "github", "repo": "contextbridge/planbridge", "path": "harnessIntegrations/claude" },
+      "description": "PlanBridge — opens plan reviews in the browser on ExitPlanMode."
     }
   ]
```

Confirmed locally against `claude plugin validate .` (Claude Code 2.1.132): the nested object form is canonical; the shorthand string `"github:..."` is rejected as invalid input.

The `cli/` directory in `contextbridge/claude-plugin` is deleted.

**Code changes in this repo:**

- `packages/cli/src/harnesses/ClaudeInstaller.ts` — `PLUGIN_ID = 'planbridge@contextbridge'`; add `LEGACY_PLUGIN_ID = 'cli@contextbridge'`. `MARKETPLACE_SOURCE` unchanged.
- `status()` surfaces both new and legacy plugin entries in `managed[]`. `installed` stays gated on the new id only.
- `runInstall(scope)`: if `cli@contextbridge` is at `scope`, uninstall it before installing the new id. Emits one stderr line about the rename.
- `runUninstall(scope)`: also uninstalls `cli@contextbridge` at the target scope when present.
- `update.ts`: after a successful binary swap, walk `ALL_INSTALLERS`, find harnesses with managed state, and spawn `<process.execPath> install <harness-id>` per harness. The new binary's installer migrates legacy and refreshes the plugin contract.
- `release-please-config.json`: add `extra-files` so `harnessIntegrations/claude/.claude-plugin/plugin.json`'s `version` bumps with each binary release.

### Transition path for existing users

`contextbridge update` now couples binary and plugin updates. After the binary swap, `runUpdate` walks installers, skips harnesses with no PlanBridge state, and spawns the new binary's per-harness `install <harness>` for those with state. For Claude with legacy `cli@contextbridge` installed, the new binary's `runInstall`:

1. detects legacy at the target scope via `claude plugin list --json`,
2. runs `claude plugin uninstall cli@contextbridge --scope <scope>`,
3. emits stderr: `PlanBridge plugin renamed from cli@contextbridge to planbridge@contextbridge — migrated automatically.`,
4. runs `claude plugin marketplace add contextbridge/claude-plugin` (idempotent),
5. runs `claude plugin install planbridge@contextbridge --scope <scope>`.

Refresh failure logs a warning and never blocks update success. `contextbridge install` continues to work as the explicit/manual entrypoint and performs the same migration. `contextbridge uninstall` cleans up the legacy plugin too if present. The Codex installer is untouched.

- **Move everything (including `marketplace.json`) into this repo and archive `contextbridge/claude-plugin`.** Considered and rejected. Keeping the marketplace repo separate is a feature, not vestigial: it's the org-level catalog of all contextbridge plugins. When the next plugin ships, it gets its own product repo and a new entry in the same `contextbridge/claude-plugin/marketplace.json`. Users see one storefront; each product owns its own implementation.
- **Keep the plugin name `cli`** to avoid the breaking change. Cheaper in the short term, but the rename only gets more expensive as the user base grows. Doing it now is the right cost-optimization.
- **Keep the split as-is.** Status quo; rejected because the constraint that drove it is gone and the drift cost grows as more harness plugins land.
- **Banner / stderr nudge for users still on the old binary.** Considered and dropped. They'll get the update eventually, and from then on every future update keeps things current.

### Additional context

Existing layout in `contextbridge/claude-plugin`:

```
.claude-plugin/marketplace.json     # marketplace "contextbridge", plugin source: "./cli"
cli/.claude-plugin/plugin.json      # plugin name "cli", version "0.1.0"
cli/hooks/hooks.json                # PermissionRequest:ExitPlanMode -> "contextbridge hook claude"
```

After this lands:

```
contextbridge/claude-plugin
└── .claude-plugin/marketplace.json     # plugin "planbridge", cross-repo source -> contextbridge/planbridge

contextbridge/planbridge
└── harnessIntegrations/
    └── claude/
        ├── .claude-plugin/plugin.json  # plugin name "planbridge"
        └── hooks/hooks.json            # unchanged contract: contextbridge hook claude
```

Cross-repo plugin source is documented at https://code.claude.com/docs/en/plugin-marketplaces.md — the docs explicitly call out that "a marketplace hosted at `acme-corp/plugin-catalog` can list a plugin fetched from `acme-corp/code-formatter`."

### Implementation

- contextbridge/claude-plugin#1 — moves the marketplace pointer to the new cross-repo source, deletes `cli/`.
- contextbridge/planbridge#52 — adds the plugin assets, renames the plugin id, couples binary and plugin updates.

**Sequencing**: merge `claude-plugin#1` first (catalog flips), then `planbridge#52`, then the planbridge release-please release PR.

blimmer:
## Summary

Move PlanBridge's Claude Code plugin assets (`plugin.json`, `hooks.json`) from `contextbridge/claude-plugin` into this repo at `harnessIntegrations/claude/`, rename the plugin id from `cli@contextbridge` to `planbridge@contextbridge`, and couple binary + plugin updates so `contextbridge update` re-runs install for harnesses that already have PlanBridge wired up. Closes #49.

## Review focus

- **`runUpdate` post-swap refresh** in `packages/cli/src/commands/update.ts` walks `ALL_INSTALLERS`, skips harnesses with `status.managed.length === 0`, and spawns `<process.execPath> install <harness-id>` per harness with state. Refresh failure logs a warning and never blocks update success. Spawning `process.execPath` (not the old `runUpdate`'s in-process logic) is deliberate — the child runs the _new_ binary's installer code, which knows about the rename and updates/migrates Claude plugin state automatically.
- **`ClaudeInstaller.status()`** now surfaces both `planbridge@contextbridge` and `cli@contextbridge` entries in `managed[]`, but `installed` stays gated on the new id only. This is what makes the refresh filter naturally migrate legacy users without a separate "do I have legacy state?" abstraction.
- **`runInstall`/`runUninstall` legacy cleanup** at the target scope only — see `ClaudeInstaller.ts`. Tests cover the cross-scope edge case (legacy at project, install at user → leaves legacy alone). Existing `planbridge@contextbridge` installs now use `claude plugin update` so `contextbridge update` pulls the latest plugin version.
- **`release-please-config.json` `extra-files`** keeps `harnessIntegrations/claude/.claude-plugin/plugin.json`'s `version` in sync with binary releases.

**Sequencing**: depends on [contextbridge/claude-plugin#1](https://github.com/contextbridge/claude-plugin/pull/1) merging first so the new plugin id resolves in the marketplace catalog. Then this PR; then the release-please release PR.

## Commits

- [`b2807dd`](https://github.com/contextbridge/planbridge/commit/b2807dda623dbbec2855ab190c1cda6656ae486f) — package the Claude Code plugin integration assets in this repo
- [`a382919`](https://github.com/contextbridge/planbridge/commit/a382919da258e6139ed01143284de55fb31bea5b) — rename the Claude plugin to `planbridge@contextbridge`
- [`9494b1f`](https://github.com/contextbridge/planbridge/commit/9494b1f00310762977a2d65cf1ee1998bbd0cf41) — refresh managed harness integrations after `contextbridge update`

### DIFF
--- a/harnessIntegrations/claude/.claude-plugin/plugin.json
+++ b/harnessIntegrations/claude/.claude-plugin/plugin.json
@@ -8,5 +8,5 @@
   "homepage": "https://plan.contextbridge.ai",
   "repository": "https://github.com/contextbridge/planbridge",
   "license": "MIT",
-  "keywords": ["contextbridge", "planbridge", "planning", "review", "hooks", "ExitPlanMode"]
+  "keywords": ["contextbridge", "planbridge", "planning", "review", "hooks"]
 }
--- a/harnessIntegrations/claude/.claude-plugin/plugin.json
+++ b/harnessIntegrations/claude/.claude-plugin/plugin.json
@@ -1,0 +1,12 @@
+{
+  "name": "planbridge",
+  "description": "PlanBridge — opens plan reviews in the browser on ExitPlanMode.",
+  "version": "0.2.0",
+  "author": {
+    "name": "ContextBridge AI, Inc."
+  },
+  "homepage": "https://plan.contextbridge.ai",
+  "repository": "https://github.com/contextbridge/planbridge",
+  "license": "MIT",
+  "keywords": ["contextbridge", "planbridge", "planning", "review", "hooks", "ExitPlanMode"]
+}
--- a/packages/cli/src/commands/install.test.ts
+++ b/packages/cli/src/commands/install.test.ts
@@ -34,7 +34,7 @@ describe('runInstall', () => {
       'user',
     ]);
     expect(pluginInstall).toHaveLength(1);
-    expect(pluginInstall[0]?.args).toEqual(['plugin', 'install', 'cli@contextbridge', '--scope', 'user']);
+    expect(pluginInstall[0]?.args).toEqual(['plugin', 'install', 'planbridge@contextbridge', '--scope', 'user']);
     expect(prompter.calls).toEqual([]);
     const stderr = io.stderr.text();
     expect(stderr).toContain('Claude Code: not installed');
@@ -49,7 +49,7 @@ describe('runInstall', () => {
       .resolves(marketplaceListResult([{ name: 'contextbridge' }]));
     commandRunner
       .on(CLAUDE_BINARY, ['plugin', 'list', '--json'])
-      .resolves(pluginListResult([{ id: 'cli@contextbridge', scope: 'user' }]));
+      .resolves(pluginListResult([{ id: 'planbridge@contextbridge', scope: 'user' }]));
 
     await runInstall(context, { yes: true });
 
@@ -57,7 +57,9 @@ describe('runInstall', () => {
     expect(commandRunner.callsTo(CLAUDE_BINARY, ['plugin', 'install'])).toEqual([]);
     expect(prompter.calls).toEqual([]);
     const stderr = io.stderr.text();
-    expect(stderr).toContain('Claude Code: installed (marketplace contextbridge; plugin cli@contextbridge @ user)');
+    expect(stderr).toContain(
+      'Claude Code: installed (marketplace contextbridge; plugin planbridge@contextbridge @ user)',
+    );
     expect(stderr).toContain('Installed 0 of 1 detected harness (1 already installed, skipped).');
   });
 
@@ -69,14 +71,14 @@ describe('runInstall', () => {
       .resolves(marketplaceListResult([{ name: 'contextbridge' }]));
     commandRunner
       .on(CLAUDE_BINARY, ['plugin', 'list', '--json'])
-      .resolves(pluginListResult([{ id: 'cli@contextbridge', scope: 'user' }]));
+      .resolves(pluginListResult([{ id: 'planbridge@contextbridge', scope: 'user' }]));
     commandRunner.on(CLAUDE_BINARY, ['plugin', 'marketplace', 'add']).resolves();
-    commandRunner.on(CLAUDE_BINARY, ['plugin', 'install']).resolves();
+    commandRunner.on(CLAUDE_BINARY, ['plugin', 'update']).resolves();
 
     await runInstall(context, { yes: true, force: true });
 
     expect(commandRunner.callsTo(CLAUDE_BINARY, ['plugin', 'marketplace', 'add'])).toHaveLength(1);
-    expect(commandRunner.callsTo(CLAUDE_BINARY, ['plugin', 'install'])).toHaveLength(1);
+    expect(commandRunner.callsTo(CLAUDE_BINARY, ['plugin', 'update'])).toHaveLength(1);
     expect(prompter.calls).toEqual([]);
     const stderr = io.stderr.text();
     expect(stderr).toContain('Installed 1 of 1 detected harness');
@@ -91,14 +93,16 @@ describe('runInstall', () => {
       .resolves(marketplaceListResult([{ name: 'contextbridge' }]));
     commandRunner
       .on(CLAUDE_BINARY, ['plugin', 'list', '--json'])
-      .resolves(pluginListResult([{ id: 'cli@contextbridge', scope: 'project' }]));
+      .resolves(pluginListResult([{ id: 'planbridge@contextbridge', scope: 'project' }]));
 
     await runInstall(context, { yes: true });
 
     expect(commandRunner.callsTo(CLAUDE_BINARY, ['plugin', 'marketplace', 'add'])).toEqual([]);
     expect(commandRunner.callsTo(CLAUDE_BINARY, ['plugin', 'install'])).toEqual([]);
     const stderr = io.stderr.text();
-    expect(stderr).toContain('Claude Code: installed (marketplace contextbridge; plugin cli@contextbridge @ project)');
+    expect(stderr).toContain(
+      'Claude Code: installed (marketplace contextbridge; plugin planbridge@contextbridge @ project)',
+    );
     expect(stderr).toContain('(1 already installed, skipped)');
   });
 
@@ -155,7 +159,7 @@ describe('runInstall', () => {
         'user',
       ]);
       expect(pluginInstall).toHaveLength(1);
-      expect(pluginInstall[0]?.args).toEqual(['plugin', 'install', 'cli@contextbridge', '--scope', 'user']);
+      expect(pluginInstall[0]?.args).toEqual(['plugin', 'install', 'planbridge@contextbridge', '--scope', 'user']);
       const stderr = io.stderr.text();
       expect(stderr).toContain('Codex CLI: status unavailable (invalid Codex hooks.json');
       expect(stderr).toContain('Installed 1 of 2 detected harnesses.');
--- a/packages/cli/src/commands/install.test.ts
+++ b/packages/cli/src/commands/install.test.ts
@@ -3,6 +3,11 @@ import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { describe, expect, it } from 'bun:test';
 import { CommanderError } from 'commander';
+import {
+  CLAUDE_MARKETPLACE_NAME,
+  CLAUDE_MARKETPLACE_SOURCE,
+  CLAUDE_PLUGIN_ID,
+} from '#src/harnesses/ClaudeInstaller.ts';
 import { getDescriptor } from '#src/harnesses/registry.ts';
 import { environment } from '#src/testFactories.ts';
 import { createStubContext, marketplaceListResult, pluginListResult } from '#src/testHelpers/index.ts';
@@ -29,12 +34,12 @@ describe('runInstall', () => {
       'plugin',
       'marketplace',
       'add',
-      'contextbridge/claude-plugin',
+      CLAUDE_MARKETPLACE_SOURCE,
       '--scope',
       'user',
     ]);
     expect(pluginInstall).toHaveLength(1);
-    expect(pluginInstall[0]?.args).toEqual(['plugin', 'install', 'planbridge@contextbridge', '--scope', 'user']);
+    expect(pluginInstall[0]?.args).toEqual(['plugin', 'install', CLAUDE_PLUGIN_ID, '--scope', 'user']);
     expect(prompter.calls).toEqual([]);
     const stderr = io.stderr.text();
     expect(stderr).toContain('Claude Code: not installed');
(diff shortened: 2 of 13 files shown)


harnessIntegrations/claude/.claude-plugin/plugin.json
```diff
@@ -0,0 +1,12 @@
+{
```
blimmer, line 1: This moves from github.com/contextbridge/claude-plugin (see https://github.com/contextbridge/claude-plugin/pull/1). That repo is now just a catalog of all ContextBridge claude plugins (right now, just PlanBridge, but could expand to other products in the future).

I'm doing this migration now, which will make https://github.com/contextbridge/planbridge/issues/51 easier to implement in a single place.

packages/cli/src/commands/install.test.ts
```diff
@@ -34,7 +34,7 @@ describe('runInstall', () => {
       'user',
     ]);
     expect(pluginInstall).toHaveLength(1);
-    expect(pluginInstall[0]?.args).toEqual(['plugin', 'install', 'cli@contextbridge', '--scope', 'user']);
+    expect(pluginInstall[0]?.args).toEqual(['plugin', 'install', 'planbridge@contextbridge', '--scope', 'user']);
```
blimmer: I'm renaming this to match up. If we don't feel strongly, we can leave it as `cli` - feedback welcome.

packages/cli/src/commands/update.test.ts
```diff
@@ -189,4 +202,193 @@ describe('runUpdate', () => {
...
+    commandRunner.on(process.execPath, ['install', 'claude']).resolves();
+
+    await runUpdate(context);
+
+    const spawnCall = commandRunner.calls.find((c) => c.cmd === process.execPath);
+    expect(spawnCall).toBeDefined();
+    expect(spawnCall?.args).toEqual(['install', 'claude', '--scope', 'user']);
+  });
+
+  it('refreshes Claude at project scope when the new plugin is installed at project scope', async () => {
```
blimmer, line 255: Things could get a little bit weird if the old named plugin is installed in various projects, but that didn't feel like something I wanted to work around.

packages/cli/src/harnesses/ClaudeInstaller.ts
```diff
@@ -42,25 +43,43 @@ export class ClaudeInstaller extends ScopedHarnessInstaller {
...
     await runPluginCommand(ctx, binaryName, 'marketplace add', [
       'marketplace',
       'add',
       MARKETPLACE_SOURCE,
       '--scope',
       scope,
     ]);
-    await runPluginCommand(ctx, binaryName, 'install', ['install', PLUGIN_ID, '--scope', scope]);
+    if (hasCurrentAtScope) {
+      await runPluginCommand(ctx, binaryName, 'update', ['update', PLUGIN_ID, '--scope', scope]);
```
blimmer: Claude will update plugins on some cadence, but we want to force it to be up-to-date (e.g., when we introduce new harness features).

release-please-config.json
```diff
@@ -6,6 +6,13 @@
       "package-name": "contextbridge",
       "include-component-in-tag": false,
       "changelog-path": "CHANGELOG.md",
+      "extra-files": [
+        {
+          "type": "json",
+          "path": "harnessIntegrations/claude/.claude-plugin/plugin.json",
+          "jsonpath": "$.version"
```
blimmer, line 13: This is a handy feature that'll keep the claude plugin released in-step with the binary.

harnessIntegrations/claude/.claude-plugin/plugin.json
```diff
@@ -0,0 +1,12 @@
...
+  "name": "planbridge",
+  "description": "PlanBridge — opens plan reviews in the browser on ExitPlanMode.",
+  "version": "0.2.0",
+  "author": {
+    "name": "ContextBridge AI, Inc."
+  },
+  "homepage": "https://plan.contextbridge.ai",
+  "repository": "https://github.com/contextbridge/planbridge",
+  "license": "MIT",
+  "keywords": ["contextbridge", "planbridge", "planning", "review", "hooks", "ExitPlanMode"]
```
jcarver989: Is `ExitPlanMode` in the keywords intentional? Just curious since it sticks out against the others.

packages/cli/src/commands/install.test.ts
jcarver989: Think `planbridge` will make more sense given the marketing site and product name

packages/cli/src/commands/install.test.ts
```diff
@@ -49,15 +49,17 @@ describe('runInstall', () => {
       .resolves(marketplaceListResult([{ name: 'contextbridge' }]));
     commandRunner
       .on(CLAUDE_BINARY, ['plugin', 'list', '--json'])
-      .resolves(pluginListResult([{ id: 'cli@contextbridge', scope: 'user' }]));
+      .resolves(pluginListResult([{ id: 'planbridge@contextbridge', scope: 'user' }]));
```
jcarver989: Seems like we should use the constant that's defined elsewhere in this PR across these tests.

packages/cli/src/commands/update.ts
```diff
@@ -57,6 +60,53 @@ export function registerUpdate(ctx: CliContext, program: Command): void {
     });
 }
 
+// After the binary has been swapped, re-run the new binary's per-harness install
+// for any harness with existing PlanBridge state. Keeps installed plugins in
+// sync with the binary across renames and hook-contract changes. Skips harnesses
+// the user never wired up; never blocks update success on a refresh failure.
+async function refreshInstalledHarnesses(ctx: CliContext): Promise<void> {
+  const { commandRunner, logger } = ctx;
+  for (const installer of ALL_INSTALLERS) {
```
jcarver989, line 68: Probably just me being me, but I'd `flatMap` here (using empty array in the failure case) as it'd avoid the `let` binding + re-assignment.

packages/cli/src/commands/update.ts
```diff
@@ -57,6 +60,53 @@ export function registerUpdate(ctx: CliContext, program: Command): void {
...
+    let refreshScopes: string[];
+    try {
+      const status = await installer.status(ctx);
+      refreshScopes = getRefreshScopes(status.managed);
+    } catch (err) {
+      logger.warn({ err, harness: installer.descriptor.id }, 'post-update harness refresh failed');
+      continue;
+    }
+
+    for (const scope of refreshScopes) {
```
jcarver989: Noting that this is sequential, assuming this is intentional since concurrent install commands might cause issues?

jcarver989 (approved):

harnessIntegrations/claude/.claude-plugin/plugin.json
blimmer: Ah, just copied this over verbatim from the other repo - you're right though, not very helpful. Let me update.

packages/cli/src/commands/update.ts
blimmer: Sure - I'll take a more functional approach throughout.

packages/cli/src/commands/update.ts
blimmer: Yep, it's intentional. `claude plugin install` mutates shared config under `~/.claude` (and Codex hook installs touch `~/.codex/hooks.json`), so parallel installs across scopes/harnesses would race on those files. Sequential is also fine performance-wise — the post-update refresh fan-out is small (one per managed scope per installer), and `stdio: 'inherit'` already streams progress to the user.

PR #53, opened 2026-05-07
refactor(testHelpers): replace FakeCommandRunner FIFO scripts with argument-keyed matchers

blimmer:
## Summary

Replaces `FakeCommandRunner.script(...)` with a fluent matcher API: `cr.on('claude', ['plugin', 'install']).resolves({ stdout: 'ok' })`. Tests now declare command outcomes by command shape instead of padding a FIFO queue, so production command ordering changes do not force unrelated test rewrites.

The final helper keeps the matcher machinery inside `FakeCommandRunner`, adds diagnostics for missed responders, and defaults `resolves()` to `{ exitCode: 0, stdout: '', stderr: '' }` with partial overrides.

## Review focus

- **API shape and defaults.** `CommandStub.resolves()` accepts `Partial<RunCommandResult>` and fills the normal successful shell result fields. `rejects()` simulates the runner promise rejecting rather than a non-zero command exit.
- **Prefix-match default for `on(cmd, args)`.** Tests can match the command prefix they care about while ignoring trailing args; `callsTo()` uses the same prefix behavior for assertions.
- **Persistent first-match responders.** Matchers are intentionally reusable and first-match-wins, which keeps repeated status queries easy to stub without duplicating FIFO entries.
- **Ordering-sensitive assertions.** `ClaudeInstaller.test.ts` still uses full ordered `commandRunner.calls` assertions where sequence is part of the contract; most other tests now assert the specific calls they care about with `.callsTo()`.

## Commits

- [`8e47d94`](https://github.com/contextbridge/planbridge/commit/8e47d94) — Add the `FakeCommandRunner` matcher API, defaulted responses, rejection support, call filtering, and focused helper tests.
- [`4670251`](https://github.com/contextbridge/planbridge/commit/4670251) — Migrate CLI, install, uninstall, Claude installer, install status, and updater tests from FIFO scripts to command matchers.

### DIFF
--- a/packages/cli/src/commands/install.test.ts
+++ b/packages/cli/src/commands/install.test.ts
@@ -5,13 +5,7 @@ import { describe, expect, it } from 'bun:test';
 import { CommanderError } from 'commander';
 import { getDescriptor } from '#src/harnesses/registry.ts';
 import { environment } from '#src/testFactories.ts';
-import {
-  type FakeCommandCall,
-  type FakeCommandRunner,
-  createStubContext,
-  marketplaceListResult,
-  pluginListResult,
-} from '#src/testHelpers/index.ts';
+import { createStubContext, marketplaceListResult, pluginListResult } from '#src/testHelpers/index.ts';
 import { runInstall } from './install.ts';
 
 const CLAUDE_BINARY = getDescriptor('claude').binaryName;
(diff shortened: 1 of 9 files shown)


packages/cli/src/commands/install.test.ts
```diff
@@ -21,23 +15,26 @@ describe('runInstall', () => {
...
-    commandRunner.script(
-      marketplaceListResult([]),
-      pluginListResult([]),
-      { exitCode: 0, stdout: '', stderr: '' },
-      { exitCode: 0, stdout: '', stderr: '' },
-    );
+    commandRunner.on(CLAUDE_BINARY, ['plugin', 'marketplace', 'list', '--json']).resolves(marketplaceListResult([]));
+    commandRunner.on(CLAUDE_BINARY, ['plugin', 'list', '--json']).resolves(pluginListResult([]));
+    commandRunner.on(CLAUDE_BINARY, ['plugin', 'marketplace', 'add']).resolves();
+    commandRunner.on(CLAUDE_BINARY, ['plugin', 'install']).resolves();
```
blimmer, line 21: I think this reads a lot nicer. As mentioned in the PR description, I was inspired by the fluent interface from aws-sdk-client-mock, which I like a lot.

packages/cli/src/testHelpers/CommandMatcherRegistry.ts
```diff
@@ -0,0 +1,86 @@
...
+
+interface RegisteredResponder {
+  readonly matcher: CommandMatcher;
+  readonly label: string;
+  readonly outcome: ResponderOutcome;
+  consumed: boolean;
+  readonly once: boolean;
+}
+
+export class CommandMatcherRegistry {
```
blimmer: OK, I admit this is a little vibey, but it didn't feel like a ton of code and the libraries I researched were a lot more heavyweight. So it felt OK to me, but feel free to push back.

packages/cli/src/testHelpers/FakeCommandRunner.test.ts
```diff
@@ -0,0 +1,179 @@
...
+      expect(cr.which('claude')).toBeNull();
+    });
+  });
+
+  // Matcher-chain semantics (registration order, resolvesOnce consumption,
+  // resolves() defaults, rejects(), miss diagnostic format) live in
+  // CommandMatcherRegistry.test.ts. Tests here cover only what's specific
+  // to FakeCommandRunner — the prefix/exact/predicate/any matcher shapes
+  // that `on()`/`onExact()`/`onMatch()`/`onAny()` build, and the runner's
+  // call bookkeeping.
```
blimmer: Review agent, remove this useless comment. Contextbot, take note - I don't want these types of comments in code. They only represent details useful during this specific agent session. Comments should only be left in code when they're not ephemeral / are useful to future developers.

blimmer:
I need this for some other work I'd like to get out today, so I'm going to merge it through. Post-merge feedback is welcome!

PR #64, opened 2026-05-08
fix: resolve contextbridge via PATH for post-update refresh

### LINKED ISSUE #63: bug: post-update harness refresh fails with ENOENT after brew upgrade
### What happened?

`contextbridge update`'s post-update harness refresh (added in #52) silently no-ops for every Homebrew-installed user because the spawn target path was deleted by brew before the spawn fires.

`runUpdate` in `packages/cli/src/commands/update.ts` calls `commandRunner.run(process.execPath, ['install', '<harness-id>', '--scope', '<scope>'])` to re-run the new binary's per-harness installer. But `process.execPath` is captured at process start — for brew-installed binaries this is the cellar path of the *pre-upgrade* version (e.g., `/opt/homebrew/Caskroom/cli@alpha/0.3.0-alpha.1/contextbridge`). Brew's `--cask` upgrade flow then runs `Purging files for version 0.3.0-alpha.1`, deleting that whole cellar directory. By the time `runUpdate` reaches the spawn, the path is gone — Node fails with `ENOENT`.

The refresh-failure logging from #52 correctly downgrades this to a warning so `update` itself still exits 0 with `✓ update complete.`, but the actual per-harness re-install never runs. The post-update refresh feature — the whole reason #52 was filed — is effectively dead for brew users.

Discovered live during alpha.1 → alpha.2 upgrade testing of #62.

### contextbridge version

0.3.0-alpha.2 (and 0.3.0-alpha.1; the failure is in the upgrade transition itself, observable on every brew-driven version bump)

### Operating system

macOS

### Steps to reproduce

1. Install contextbridge via the brew tap (e.g., `brew install contextbridge/tap/cli@alpha`).
2. Run `contextbridge install claude --yes` so a managed harness exists for the refresh to act on.
3. Wait for (or simulate) a newer release on the same channel.
4. Run `LOG_LEVEL=trace contextbridge update`.
5. Observe `ENOENT` warnings for each managed harness while update exits 0.

After the binary swap, `runUpdate` re-runs the new binary's `install <harness>` for each managed scope. Trace should show the per-harness install shellouts firing against the post-upgrade binary path.

### What actually happened?

Brew swaps the binary and purges the old cellar directory. `runUpdate` then tries to `posix_spawn` against the captured-at-startup `process.execPath` — which is the now-deleted old path. Each managed harness logs `post-update harness refresh failed` with `ENOENT`. Update reports success and exits 0, but no refresh happens.

### Relevant logs or stack trace

```shell
$ LOG_LEVEL=trace contextbridge update
A new version is available: v0.3.0-alpha.2 (you're on v0.3.0-alpha.1).
...
==> Upgrading cli@alpha
  0.3.0-alpha.1 -> 0.3.0-alpha.2
==> Unlinking Binary '/opt/homebrew/bin/contextbridge'
==> Linking Binary 'contextbridge' to '/opt/homebrew/bin/contextbridge'
==> Purging files for version 0.3.0-alpha.1 of Cask cli@alpha
🍺  cli@alpha was successfully upgraded!

{"level":40,"msg":"post-update harness refresh failed","err":{"type":"Error","message":"ENOENT: no such file or directory, posix_spawn '/opt/homebrew/Caskroom/cli@alpha/0.3.0-alpha.1/contextbridge'","code":"ENOENT","path":"/opt/homebrew/Caskroom/cli@alpha/0.3.0-alpha.1/contextbridge","syscall":"posix_spawn","errno":-2},"harness":"claude","scope":"user"}
{"level":40,"msg":"post-update harness refresh failed","err":{"type":"Error","message":"ENOENT: no such file or directory, posix_spawn '/opt/homebrew/Caskroom/cli@alpha/0.3.0-alpha.1/contextbridge'","code":"ENOENT","path":"/opt/homebrew/Caskroom/cli@alpha/0.3.0-alpha.1/contextbridge","syscall":"posix_spawn","errno":-2},"harness":"codex","scope":"user"}
✓ update complete.
```

`which contextbridge` immediately after returns `/opt/homebrew/bin/contextbridge` (the stable symlink), which `readlink` resolves to `…/0.3.0-alpha.2/contextbridge`. The new binary is in place and would have worked if it had been the spawn target.

### Anything else

- Discovered while testing #62 (the marketplace-cache-refresh fix). #62 itself is unaffected by this bug — they're separate code paths. Linking for context.
- Introduced in #52 (`refresh plugins on update`) — `runUpdate` walks `ALL_INSTALLERS` and spawns `process.execPath install <harness-id>` for each managed scope.
- Why the unit tests in `packages/cli/src/commands/update.test.ts` don't catch this: `FakeCommandRunner` happily resolves stubs registered against `process.execPath`, so `executed → refresh fires → install spawns` looks fine in tests. The cellar-purge happens outside the test process.
- Affected scope: every brew-installed user (likely the majority of macOS users).
- Install-script users may be unaffected if the install script writes the binary to a stable path that survives upgrades — worth verifying, but the brew path is definitely broken.

blimmer:
## Summary

Fixes #63. After `brew upgrade --cask cli@alpha` (or `cli`), the post-update harness refresh introduced in #52 fails silently with `ENOENT` because `process.execPath` was captured at process start and now points at the cellar dir brew already purged. The refresh exists to keep harness state in sync across binary changes (renames, hook contracts) — silent no-op defeats the entire feature for every brew-installed user.

Resolve the binary via `commandRunner.which('contextbridge')` instead. The PATH symlink (`/opt/homebrew/bin/contextbridge` for brew, `~/.local/bin/contextbridge` for install-script users) survives the upgrade. If `which` returns null, log `error` and skip — refusing to fall back to the broken `execPath` keeps us from silently masking the very bug we're fixing.

Also bumped refresh-failure logs from `warn` → `error` so Sentry surfaces these in production.

## Review focus

- **Logging level decision** in `update.ts`: I bumped four warn paths to error (which-null, status-throws, exitCode-non-zero, spawn-throws). All four are anomalies for a working install; a non-actionable user environment issue should not realistically trigger them. If you'd rather a subset stay at warn, let me know.
- **No fallback to `execPath`** when `which` returns null. Considered an `existsSync(process.execPath)` fallback for users with non-PATH installs but rejected: such users are rare (both supported install methods put the binary on PATH), and the fallback would mask exactly the brew-purge case the fix targets.
- **Test coverage**: the seven existing refresh tests were updated to stub against the PATH-resolved path and never stub `execPath`, so any future change reintroducing `execPath` spawning fails them via `FakeCommandRunner: no responder`. One new test covers the `which`-null early-return branch (the only genuinely new behavior).

## Commits

- [`ae70910`](https://github.com/contextbridge/planbridge/commit/ae70910a7a541e37e1d61a2859850a5ce40e19db) — replace `process.execPath` with PATH-resolved binary in post-update refresh; bump warn → error on refresh failures.

### DIFF
--- a/packages/cli/src/commands/update.ts
+++ b/packages/cli/src/commands/update.ts
@@ -65,9 +65,16 @@ export function registerUpdate(ctx: CliContext, program: Command): void {
 // sync with the binary across renames and hook-contract changes. Skips harnesses
 // the user never wired up; never blocks update success on a refresh failure.
 async function refreshInstalledHarnesses(ctx: CliContext): Promise<void> {
+  const { commandRunner, logger } = ctx;
+  // Resolve via PATH, not process.execPath: brew purges the old cellar dir before this runs.
+  const binaryPath = commandRunner.which('contextbridge');
+  if (!binaryPath) {
+    logger.error('post-update harness refresh skipped: contextbridge not found on PATH');
+    return;
+  }
   for (const installer of ALL_INSTALLERS) {
     for (const scope of await getInstallerRefreshScopes(ctx, installer)) {
-      await refreshInstallerScope(ctx, installer, scope);
+      await refreshInstallerScope(ctx, binaryPath, installer, scope);
     }
   }
 }
@@ -77,25 +84,30 @@ async function getInstallerRefreshScopes(ctx: CliContext, installer: HarnessInst
     const status = await installer.status(ctx);
     return getRefreshScopes(status.managed);
   } catch (err) {
-    ctx.logger.warn({ err, harness: installer.descriptor.id }, 'post-update harness refresh failed');
+    ctx.logger.error({ err, harness: installer.descriptor.id }, 'post-update harness refresh failed');
     return [];
   }
 }
 
-async function refreshInstallerScope(ctx: CliContext, installer: HarnessInstaller, scope: string): Promise<void> {
+async function refreshInstallerScope(
+  ctx: CliContext,
+  binaryPath: string,
+  installer: HarnessInstaller,
+  scope: string,
+): Promise<void> {
   const { commandRunner, logger } = ctx;
   try {
-    const result = await commandRunner.run(process.execPath, ['install', installer.descriptor.id, '--scope', scope], {
+    const result = await commandRunner.run(binaryPath, ['install', installer.descriptor.id, '--scope', scope], {
       stdio: 'inherit',
     });
     if (result.exitCode !== 0) {
-      logger.warn(
+      logger.error(
         { exitCode: result.exitCode, harness: installer.descriptor.id, scope },
         'post-update harness refresh failed',
       );
     }
   } catch (err) {
-    logger.warn({ err, harness: installer.descriptor.id, scope }, 'post-update harness refresh failed');
+    logger.error({ err, harness: installer.descriptor.id, scope }, 'post-update harness refresh failed');
   }
 }
 
--- a/packages/cli/src/commands/update.test.ts
+++ b/packages/cli/src/commands/update.test.ts
@@ -16,6 +16,9 @@ import { runUpdate } from './update.ts';
 
 const CLAUDE_BINARY = getDescriptor('claude').binaryName;
 const CODEX_BINARY = getDescriptor('codex').binaryName;
+// Arbitrary fake path returned from `commandRunner.which('contextbridge')` so tests
+// can stub the refresh-spawn target separately from process.execPath.
+const FAKE_CONTEXTBRIDGE_PATH = '/usr/local/bin/contextbridge';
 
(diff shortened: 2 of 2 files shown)


packages/cli/src/commands/update.ts
```diff
@@ -77,25 +84,30 @@ async function getInstallerRefreshScopes(ctx: CliContext, installer: HarnessInst
     const status = await installer.status(ctx);
     return getRefreshScopes(status.managed);
   } catch (err) {
-    ctx.logger.warn({ err, harness: installer.descriptor.id }, 'post-update harness refresh failed');
+    ctx.logger.error({ err, harness: installer.descriptor.id }, 'post-update harness refresh failed');
```
blimmer, line 87: I want to know about this in sentry.

Issue #55, opened 2026-05-08
feat: Evaluate and package PlanBridge as a Codex plugin


blimmer:
### What's the problem?

PlanBridge currently installs its Codex integration by writing Codex hook config directly. Codex now has a plugin system that can package lifecycle hooks via a plugin manifest and `hooks/hooks.json`, which may be a better distribution model for PlanBridge once the plugin hook path is stable enough.

This should be tracked separately from the immediate Codex v0.129 feature-flag fix so the deprecation warning and hook trust docs can be addressed without blocking on plugin design.

### Proposed solution

Evaluate and package PlanBridge as a Codex plugin in a separate implementation path:

* Confirm current Codex plugin hook behavior across current/latest Codex releases.
* Verify whether plugin-bundled hooks still require the separate `plugin_hooks` feature flag, and document the exact requirement.
* Design a PlanBridge plugin package with `.codex-plugin/plugin.json` and `hooks/hooks.json` that runs `contextbridge hook codex` for the Codex `Stop` lifecycle event.
* Decide whether the plugin should eventually replace direct `hooks.json` installs or remain an optional/experimental path.
* Document marketplace setup, install/enable flow, feature flags, and hook trust behavior.

### Alternatives you considered

Keep direct `hooks.json` installation as the only Codex integration. That remains the right immediate path for issue contextbridge/planbridge#54, but plugin packaging may provide a cleaner long-term Codex distribution story.

### Additional context

Relevant findings from local Codex v0.129.0 exploration:

* Codex docs say plugins can ship lifecycle config via `hooks/hooks.json` or a manifest `hooks` field.
* `codex features list` in v0.129.0 shows `plugins` and `hooks` as stable/default true.
* `plugin_hooks` appears as under-development/default false in v0.129.0.
* Source/string inspection indicates plugin-bundled hook loading is distinct from direct user/project `hooks.json` loading.
* This issue should be implemented after the immediate v0.129 fix lands.

Related: contextbridge/planbridge#54

PR #83, opened 2026-05-12
feat: add 'contextbridge open' subcommand and planbridge-open skill

### LINKED ISSUE #51: feat: skill to open an arbitrary file in PlanBridge

blimmer:
## Summary

Adds a new `contextbridge open` CLI subcommand and a `planbridge-open` skill that opens an arbitrary markdown file (or piped content) in the PlanBridge browser UI for human annotation. The skill ships to Claude via the existing plugin and to Codex via a small registry-driven installer — adding a future Claude skill is now a one-line registration; the Codex side picks it up automatically. The user invokes the skill, the LLM resolves the path, calls the CLI, and surfaces the human's annotations on stdout as markdown for the LLM to act on conversationally. Closes #51.

## Review focus

- **Engine boundary extension** — schema gains `'document'` ContentKind, `'open_command'` entrypoint, and optional `metadata.sourcePath`. `runAnnotation` and `formatAgentResponse` thread `sourcePath` through; payload metadata also drives the frontend page title (basename of the source path when present). The plan flow is byte-identical to before — PLAN_TEMPLATES callers don't pass `sourcePath` and the plan `.hbs` files never reference `{{source}}`.
- **Shared `abort` helper** — the per-command `abort()` in `plan.ts` and `open.ts` was byte-for-byte identical; extracted to `packages/cli/src/commands/abort.ts`. `hookClaude.ts` and `hookCodex.ts` have a different abort contract (always log at error level) and were intentionally NOT consolidated.
- **Shared test helpers** — `createDeferred<T>` and the rich tracked `createAnnotationDependencies` stub were duplicated across `runAnnotation.test.ts`, `plan.test.ts`, and `open.test.ts`. Both are now exported from `testHelpers/`. The hook tests' existing `createAnnotationDependencies({ submission })` calls continue to work — the richer return type is assignable to the bare `AnnotationDependencies` interface.
- **Skill registry & Codex auto-distribution** — `packages/cli/src/bundledSkills.ts` exports a `bundledSkills` array of `{ name, body }` entries; the Claude SKILL.md is the single source of truth, with the embedded copy validated by a per-skill sync test. CodexInstaller iterates over this registry for install / uninstall / status, so adding a new Claude skill requires zero installer changes (just create the SKILL.md and add one line to the registry). A `.claude/rules/claude-plugin-skills.md` rule (auto-loaded when editing `SKILL.md` or `bundledSkills.ts`) documents the workflow.
- **Codex install layout** — skills go to `~/.agents/skills/planbridge-<name>/SKILL.md` (cross-harness convention also used by crit/OpenCode/Pi/etc.). Skills are always installed at user level regardless of hook scope. User-scope uninstall removes only `planbridge-<name>/`; project-scope uninstall leaves them alone. The installer touches only its own subdirectories under the shared `.agents/skills/` tree.
- **UI generalization** — visible strings became neutral (`Loading…`, `Approved.`, `Review — PlanBridge`); internals (CSS classes, React state, analytics events) intentionally unchanged to avoid PostHog dashboard churn. Chromatic baselines need re-approval for the string changes.
- **Codex plugin marketplace distribution** is intentionally deferred — `plugin_hooks` is `Stage::UnderDevelopment` in Codex, so adopting plugins only for skills would split the install model. Tracked in #55.

## Commits

- [`ca6e2ce`](https://github.com/contextbridge/planbridge/commit/ca6e2ce) — feat(cli): extend annotation engine for 'document' contentKind with sourcePath plumbing (includes shared test helpers)
- [`7906a78`](https://github.com/contextbridge/planbridge/commit/7906a78) — feat(cli): add 'contextbridge open' subcommand (includes shared abort helper and the plan/open test refactor to shared helpers)
- [`a946e5a`](https://github.com/contextbridge/planbridge/commit/a946e5a) — feat(annotation): generalize UI copy and surface sourcePath-aware title
- [`b939586`](https://github.com/contextbridge/planbridge/commit/b939586) — feat(harness): ship Claude-plugin skills via Codex installer (registry + auto-loaded rule for the skill-addition workflow)
- [`2a19fd5`](https://github.com/contextbridge/planbridge/commit/2a19fd5) — docs(cli): note 'contextbridge open' as a sibling entrypoint

### DIFF
--- a/harnessIntegrations/claude/skills/open/SKILL.md
+++ b/harnessIntegrations/claude/skills/open/SKILL.md
@@ -1,0 +1,56 @@
+---
+name: open
+description: Open a markdown file, document, or piece of content in the PlanBridge browser UI for human annotation. Use when the user wants to give human feedback on a file, draft, plan, spec, or any markdown content.
+---
+
+# Open content in PlanBridge
+
+Opens a markdown file or piped content in the PlanBridge browser UI, where the user can leave inline annotations and general comments. Their feedback is returned on stdout.
+
+## When to use
+
+Use this skill when the user wants to annotate something themselves in the browser. Typical triggers:
+
+- "I want to annotate this in PlanBridge"
+- "Open this in the browser so I can mark it up"
+- "Let me leave comments on this before you continue"
+- The user asks you to revise based on their annotations and you haven't collected them yet
+
+## How to invoke
+
+The CLI takes a file path or stdin content:
+
+    # A path on disk (most common)
+    contextbridge open /absolute/path/to/file.md
+
+    # Piped content (when the content lives in conversation, not on disk)
+    printf %s "<content>" | contextbridge open
+
+### Resolving the argument
+
+The user's argument may be a literal path or a human-language description. Resolve it first:
+
+1. If it looks like a path and the file exists, use it directly.
+2. If it's a description (e.g., "the plan we discussed", "the spec for X"), use your normal file-discovery tools to locate the file in the working directory or recent conversation, then call `contextbridge open <resolved-path>`.
+3. If the content lives in conversation rather than on disk (e.g., the user wants to annotate something you just wrote), pipe it via stdin.
+
+If you can't resolve the argument confidently, ask the user for clarification before running anything.
+
+## What happens
+
+1. PlanBridge starts a local browser session and prints the URL.
+2. The user annotates in the browser. Block on the CLI until they submit.
+3. The CLI prints a markdown summary of the user's feedback to stdout.
+
+## What to do with the output
+
+The output is feedback for discussion, not a directive. Read it carefully:
+
+- If the user left no annotations, acknowledge briefly and continue with whatever they were doing.
+- If the user left annotations, respond conversationally. They may want edits, may want discussion, may be flagging things for later. Don't assume edits are required unless the comments make that clear. When in doubt, ask what they want to do next.
+
+Treat the comments the way you'd treat a colleague's review notes — context for the next step, not a checklist to silently execute.
+
+## Limitations
+
+Folders, URLs, and multi-file annotation are not currently supported. Pass a single file path or pipe a single document.
--- a/packages/annotation/src/App.test.tsx
+++ b/packages/annotation/src/App.test.tsx
@@ -43,7 +43,7 @@ describe('App', () => {
   it('shows an empty state when the plan content is blank', () => {
     renderApp({ initialPayload: { contentKind: 'plan', content: '' } });
 
-    expect(screen.getByTestId(appTestIds.emptyState)).toHaveTextContent('No plan content was provided.');
+    expect(screen.getByTestId(appTestIds.emptyState)).toHaveTextContent('No content was provided.');
   });
 
   it('renders the update-notice card when /update-notice resolves with a notice', async () => {
@@ -292,7 +292,32 @@ describe('App', () => {
     });
 
     await waitFor(() => {
-      expect(document.title).toBe('Plan Review — PlanBridge');
+      expect(document.title).toBe('Review — PlanBridge');
+    });
+  });
+
+  it('uses the basename of metadata.sourcePath in the document title when present', async () => {
+    const payload: AnnotationPayload = {
+      content: '# A doc\n',
+      contentKind: 'document',
+      metadata: { entrypoint: 'open_command', sourcePath: '/abs/path/to/draft.md' },
+    };
+    renderApp({ initialPayload: payload });
+    await waitFor(() => {
+      expect(document.title).toBe('draft.md — PlanBridge');
+    });
+  });
+
+  it('falls back to payload title when no sourcePath is set', async () => {
+    const payload: AnnotationPayload = {
+      content: '# A doc\n',
+      contentKind: 'document',
+      title: 'A doc',
+      metadata: { entrypoint: 'open_command' },
+    };
+    renderApp({ initialPayload: payload });
+    await waitFor(() => {
+      expect(document.title).toBe('A doc — PlanBridge');
     });
   });
 
--- a/packages/cli/src/annotation/runAnnotation.test.ts
+++ b/packages/cli/src/annotation/runAnnotation.test.ts
@@ -1,9 +1,8 @@
-import type { AnnotationPayload, AnnotationSubmission } from '@contextbridge/shared/annotationSchema';
-import { annotationSubmission } from '@contextbridge/shared/testFactories';
+import type { AnnotationSubmission } from '@contextbridge/shared/annotationSchema';
 import { describe, expect, it } from 'bun:test';
 import { annotationArgs } from '#src/testFactories.ts';
-import { createStubContext } from '#src/testHelpers/index.ts';
-import { type AnnotationDependencies, runAnnotation } from './runAnnotation.ts';
+import { createAnnotationDependencies, createDeferred, createStubContext } from '#src/testHelpers/index.ts';
+import { runAnnotation } from './runAnnotation.ts';
 
 describe('runAnnotation', () => {
   it('opens the browser and returns the submitted review', async () => {
@@ -63,80 +62,26 @@ describe('runAnnotation', () => {
     expect(deps.sigintHandlerRemoved).toBe(true);
     expect(analytics.captures.some((c) => c.event === 'plan_review_submitted')).toBe(false);
   });
-});
 
-function createAnnotationDependencies(
-  options: {
-    result?: Promise<AnnotationSubmission>;
-  } = {},
-): AnnotationDependencies & {
-  closeCount: number;
-  payloads: AnnotationPayload[];
-  sigintHandlerRegistered: Promise<void>;
-  sigintHandlerRemoved: boolean;
-  submission: AnnotationSubmission;
-  triggerSigint(): void;
-} {
-  const payloads: AnnotationPayload[] = [];
-  const submission = annotationSubmission.build();
-  const sigintRegistration = createDeferred<void>();
-  let closeCount = 0;
-  let sigintHandler: (() => void) | null = null;
-  let sigintHandlerRemoved = false;
-
-  return {
-    get closeCount() {
-      return closeCount;
-    },
-    payloads,
-    sigintHandlerRegistered: sigintRegistration.promise,
-    get sigintHandlerRemoved() {
-      return sigintHandlerRemoved;
-    },
-    submission,
-    triggerSigint() {
-      if (!sigintHandler) {
-        throw new Error('SIGINT handler was not registered');
-      }
-
-      sigintHandler();
-    },
-    loadHtml: () => Promise.resolve('<html><body>annotation</body></html>'),
-    startReviewServer: (_ctx, { payload }) => {
-      payloads.push(payload);
-      return {
-        port: 4312,
-        url: 'http://localhost:4312',
-        result: options.result ?? Promise.resolve(submission),
-        close: () => {
-          closeCount += 1;
-          return Promise.resolve();
-        },
-      };
-    },
-    registerSigintHandler: (handler) => {
-      sigintHandler = handler;
-      sigintHandlerRemoved = false;
-      sigintRegistration.resolve();
-      return () => {
-        sigintHandlerRemoved = true;
-        sigintHandler = null;
-      };
-    },
-  };
-}
-
-function createDeferred<T>(): {
-  promise: Promise<T>;
-  resolve: (value: T | PromiseLike<T>) => void;
-  reject: (reason?: unknown) => void;
-} {
-  let resolve!: (value: T | PromiseLike<T>) => void;
-  let reject!: (reason?: unknown) => void;
-  const promise = new Promise<T>((promiseResolve, promiseReject) => {
-    resolve = promiseResolve;
-    reject = promiseReject;
+  it('places sourcePath in payload.metadata.sourcePath when provided', () => {
+    const { context } = createStubContext();
+    const deps = createAnnotationDependencies();
+    expect(
+      runAnnotation(
+        context,
+        { content: '# doc', contentKind: 'document', entrypoint: 'open_command', sourcePath: '/abs/doc.md' },
+        deps,
+      ),
+    ).resolves.toEqual(deps.submission);
+    expect(deps.payloads[0]?.metadata?.sourcePath).toBe('/abs/doc.md');
   });
 
-  return { promise, resolve, reject };
-}
+  it('omits sourcePath from payload.metadata when not provided', () => {
+    const { context } = createStubContext();
+    const deps = createAnnotationDependencies();
+    expect(
+      runAnnotation(context, { content: '# doc', contentKind: 'document', entrypoint: 'open_command' }, deps),
+    ).resolves.toEqual(deps.submission);
+    expect(deps.payloads[0]?.metadata?.sourcePath).toBeUndefined();
+  });
+});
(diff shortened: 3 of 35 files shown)


packages/cli/src/annotation/runAnnotation.ts
```diff
@@ -24,6 +24,7 @@ export interface RunAnnotationArgs {
   content: string;
   contentKind: ContentKind;
   entrypoint: AnnotationEntrypoint;
+  sourcePath?: string;
```
blimmer, line 27: Plans from hooks don't always have `source` (they might be stdin via the plan hook), so this is optional.

packages/cli/src/formatters/document/templates/annotationSection.hbs
```diff
@@ -0,0 +1,13 @@
+## Comment ({{range}})
```
blimmer, line 1: These are the new generic "document" templates - slightly different than `plan`.

packages/cli/src/commands/plan.ts
```diff
@@ -63,15 +65,3 @@ export function registerPlan(ctx: CliContext, program: Command): void {
       await runPlan(ctx, { path });
     });
 }
-
-function abort(ctx: CliContext, kind: 'input' | 'runtime', message: string): never {
```
blimmer, line 67: DRYing this out a bit.

packages/annotation/src/App.test.tsx
```diff
@@ -43,7 +43,7 @@ describe('App', () => {
   it('shows an empty state when the plan content is blank', () => {
```
blimmer, line 43: This commit just removes the default "Plan" language used throughout.

packages/cli/src/harnesses/CodexInstaller.ts
```diff
@@ -1,10 +1,11 @@
-import { mkdir, readFile, writeFile } from 'node:fs/promises';
```
blimmer, line 1: Codex plugins can ship skills and have an experimental flag for hooks. However, the ecosystem is pretty new so I'm just following our old pattern of writing the file directly. We can improve this in the future once codex plugins stabilize (esp wrt hooks - see #55)

packages/cli/src/harnesses/CodexInstaller.ts
```diff
@@ -19,6 +20,8 @@ const CODEX_HOOK_STATUS_MESSAGE = 'Opening PlanBridge';
 const MINIMUM_CODEX_VERSION = '0.129.0';
 
 const STOP_HOOK_KEY = 'Stop';
+const SKILL_ID = 'planbridge-open';
+const AGENTS_SKILLS_DIR = '.agents/skills';
```
blimmer, line 24: Many harnesses read from this directory so we'll also get some transparent support for other harnesses from this too.

packages/cli/src/bundledSkills.ts
```diff
@@ -0,0 +1,8 @@
+// Embedded skill content. The Codex installer writes this to disk under
+// `~/.agents/skills/planbridge-open/SKILL.md`; the Claude side ships the same
+// file as a plugin asset (`harnessIntegrations/claude/skills/open/SKILL.md`).
+// A sync test (`bundledSkills.test.ts`) asserts the embedded copy stays in
+// sync with the on-disk Claude source.
+import openSkill from '../../../harnessIntegrations/claude/skills/open/SKILL.md' with { type: 'text' };
```
blimmer, line 11: It felt like just pulling the skill into the built binary was easiest / best.

packages/cli/src/annotation/runAnnotation.test.ts
```diff
@@ -63,80 +62,26 @@ describe('runAnnotation', () => {
     expect(deps.sigintHandlerRemoved).toBe(true);
     expect(analytics.captures.some((c) => c.event === 'plan_review_submitted')).toBe(false);
   });
-});
-
-function createAnnotationDependencies(
```
blimmer, line 68: hoisting for reuse

packages/cli/src/commands/abort.ts
```diff
@@ -0,0 +1,18 @@
+import { CommanderError } from 'commander';
+import type { CliContext } from '#src/context.ts';
+
+/**
+ * Log and throw a CommanderError for a user-recoverable ('input') or runtime
+ * failure inside a subcommand handler.
+ */
+export function abort(ctx: CliContext, command: string, kind: 'input' | 'runtime', message: string): never {
```
blimmer, line 8: Again, hoisting for reuse.

packages/annotation/src/App.test.tsx
blimmer, line 40: Just a few UI tweaks to make the UI not plan-specific.

packages/cli/src/bundledSkills.ts
```diff
@@ -0,0 +1,20 @@
...
+// writes each entry to disk under `~/.agents/skills/planbridge-<name>/SKILL.md`;
+// the Claude side ships the same files as plugin assets at
+// `harnessIntegrations/claude/skills/<name>/SKILL.md`. A sync test
+// (`bundledSkills.test.ts`) asserts each embedded copy stays in lockstep with
+// its on-disk Claude source.
+//
+// To add a new skill: drop `harnessIntegrations/claude/skills/<name>/SKILL.md`
+// in place, then add a single line to the `bundledSkills` array below — the
+// Codex installer iterates over this registry and writes each one.
+import openSkill from '../../../harnessIntegrations/claude/skills/open/SKILL.md' with { type: 'text' };
```
blimmer, line 11: I'm just reusing the claude directory for now. Was thinking we can revisit this when/if we support more than claude/cursor. Felt simplest to do.

packages/cli/src/formatters/document/templates/changesRequested.hbs
```diff
@@ -0,0 +1,5 @@
+# Review submitted
+
+The human reviewed {{#if source}}`{{source}}`{{else}}the content{{/if}} and left the following comments. Read the comments and respond to the human conversationally — they may want edits, may want discussion, may be flagging things for later. Do not assume edits are required unless the human's comments make that clear.
```
jcarver989, line 3: In LLM parlance the term "user" is typically used for the human and "assistant" for the AI, this is reflected in training data and APIs so might be worth find/replacing at some point (but also orthogonal to this PR)

packages/cli/src/formatters/document/markdown.test.ts
```diff
@@ -0,0 +1,114 @@
+import type { AnnotationSubmission } from '@contextbridge/shared/annotationSchema';
+import { annotationAnchor, annotationThread, commentMessage, globalThread } from '@contextbridge/shared/testFactories';
+import { describe, expect, it } from 'bun:test';
+import { formatAgentResponse } from '#src/formatters/annotation/markdown.ts';
+import { DOCUMENT_TEMPLATES } from './templates.ts';
+
+describe('formatAgentResponse with DOCUMENT_TEMPLATES', () => {
+  const content = '# Heading\n\nSome paragraph.\n\nAnother paragraph.\n';
```
jcarver989: Can just use a multi line string, or an array with join here vs literal new lines.

packages/cli/src/commands/abort.ts
jcarver989, line 8: I'd be tempted to make `CommanderError` take these parameters in the constructor, or a static factory method and just have the caller `throw` , that composes better with `neverthrow` as you could construct the error and pass it through a `Result`.

packages/cli/src/commands/open.test.ts
```diff
@@ -0,0 +1,120 @@
...
+describe('open handler', () => {
+  let tmp: string;
+
+  beforeEach(() => {
+    tmp = mkdtempSync(join(tmpdir(), 'cb-open-test-'));
+  });
+
+  afterEach(() => {
+    rmSync(tmp, { recursive: true, force: true });
+  });
```
jcarver989, line 28: Is this safe if you're running tests concurrently? It looks like it might be since you're creating a fresh tmp dir and the `tmp` assignment occurs within a single thread.

packages/cli/src/commands/open.test.ts
```diff
@@ -0,0 +1,120 @@
...
+    const deps = createAnnotationDependencies({ submission });
+    io.stdin.write('# From stdin\n');
+    io.stdin.end();
+
+    await runOpen(context, {}, deps);
+
+    expect(io.stdout.text()).toBe(formatAgentResponse(DOCUMENT_TEMPLATES, submission, deps.payloads[0]!.content));
+    expect(deps.payloads[0]?.contentKind).toBe('document');
+    expect(deps.payloads[0]?.metadata?.entrypoint).toBe('open_command');
+    expect(deps.payloads[0]?.metadata?.sourcePath).toBeUndefined();
```
jcarver989: would be nicer to: 

```typescript
const [payload] = deps.payload;
expect(payload).toMatchObject(...)
```

packages/cli/src/commands/open.test.ts
```diff
@@ -0,0 +1,120 @@
...
+    );
+  });
+
+  it('resolves a relative path to absolute for sourcePath', async () => {
+    const file = join(tmp, 'rel.md');
+    writeFileSync(file, '# rel\n');
+
+    const { context, io } = createStubContext();
+    const deps = createAnnotationDependencies();
+    io.stdin.isTTY = true;
```
jcarver989, line 72: If tests are running concurrently is it safe to mutate this?

packages/cli/src/commands/open.ts
```diff
@@ -0,0 +1,74 @@
...
+import { formatAgentResponse } from '#src/formatters/annotation/markdown.ts';
+import { DOCUMENT_TEMPLATES } from '#src/formatters/document/templates.ts';
+import { readStreamToString } from '#src/streams.ts';
+import { abort } from './abort.ts';
+
+export interface OpenArgs {
+  path?: string;
+}
+
+export async function runOpen(ctx: CliContext, args: OpenArgs, deps?: AnnotationDependencies): Promise<void> {
```
jcarver989, line 19: It looks like we have a fair amount of coupling to the `io` module here and in tests, it seems like it'd be nicer to have a `Writer` and/or `Reader` interface to abstract that and allow for using a fake in the tests.

harnessIntegrations/claude/skills/open/SKILL.md
```diff
@@ -0,0 +1,56 @@
+---
+name: open
```
jcarver989, line 2: Do you want the agent to be able to invoke this or only the user? This looks like it's written for both the agent to invoke and the human (and claude iirc will default to both)

jcarver989 (approved):

packages/cli/src/formatters/document/templates/changesRequested.hbs
blimmer: Will do in a follow-up. We're using this in the other prompt, too. clanker - /open-github-issue for this.

packages/cli/src/commands/open.test.ts
blimmer: If this works like jest / vitest, this should be fine. Those runners run the tests serially within a single file. Review agent - please verify.

packages/cli/src/commands/open.test.ts
blimmer: This is using a FakeIo and, similar to above, I'm pretty sure we're good because the tests in this file should be run serially.

harnessIntegrations/claude/skills/open/SKILL.md
blimmer: I think both. That way, the user can say something like "let me review this in the browser" and it could pipe arbitrary text into the `open` command.

packages/cli/src/formatters/document/templates/changesRequested.hbs
blimmer: #84

packages/cli/src/commands/open.test.ts
blimmer: Yep, this is fine. Confirmed that `bun test` runs tests within the file serially. And `mkdtempSync` adds a random suffix so I think we'd be OK even with concurrency.

packages/cli/src/commands/open.ts
blimmer: I agree that we could potentially improve this in the Io module we've got. Right now, it's just a thin wrapper and exposes a lot. We could rework our Io module to be more abstracted. Clanker - /open-github-issue for this.

blimmer: #85

packages/cli/src/commands/abort.ts
blimmer: If we want to go with neverthrow for these CLI methods, we should do it to all of them. Filed #86

PR #91, opened 2026-05-12
refactor: narrow cli io surface

blimmer:
## Summary

Narrow the handler-facing CLI IO contract so command code depends on explicit stdin/stdout/stderr helpers instead of the full Node stream surface. Keep full streams available at the library-glue boundaries for pino, clack, and child-process wiring.

## Review focus

The main trade-off is that createClackPrompter accepts the concrete IoImpl because clack needs real stream objects, while CliContext.io exposes only the narrow public Io interface to handlers. FakeIo remains stream-backed by extending IoImpl, which avoids duplicate adapter behavior while preserving realistic tests.

## Commits

- [363cf0c](https://github.com/contextbridge/planbridge/commit/363cf0cba6ef1e8ceec3b54ded7b5c42eb5ad59e) — narrow the CLI IO surface and simplify the fake

### DIFF
--- a/packages/cli/src/IoImpl.ts
+++ b/packages/cli/src/IoImpl.ts
@@ -1,24 +1,58 @@
 import type { Readable, Writable } from 'node:stream';
-import type { Public } from '@contextbridge/shared/types';
+import { readStreamToString } from '#src/streams.ts';
 
-export type Io = Public<IoImpl>;
+export interface Io {
+  readonly stdinIsTTY?: boolean;
+  readStdin(): Promise<string>;
+  writeStdout(chunk: string): void;
+  writeStderr(chunk: string): void;
+}
 
-export class IoImpl {
+type IoStreams = {
+  readonly stdout: Writable & Writer;
+  readonly stderr: Writable & Writer;
+  readonly stdin: Readable & Reader;
+};
+
+export interface Writer {
+  write(chunk: string): boolean | void;
+  readonly isTTY?: boolean;
+}
+
+export interface Reader extends AsyncIterable<string | Buffer> {
+  readonly isTTY?: boolean;
+}
+
+export class IoImpl implements Io, IoStreams {
   readonly stdout: Writable & { isTTY?: boolean };
   readonly stderr: Writable & { isTTY?: boolean };
   readonly stdin: Readable & { isTTY?: boolean };
 
-  constructor() {
+  constructor(streams: Partial<IoStreams> = {}) {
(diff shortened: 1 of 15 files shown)


packages/cli/src/prompter.ts
```diff
@@ -34,7 +34,7 @@ export interface Prompter {
   select<T extends SelectValue>(options: SelectOptions<T>): Promise<T>;
 }
 
-export function createClackPrompter(io: Io): Prompter {
+export function createClackPrompter(io: IoImpl): Prompter {
```
blimmer: Clack requires the full `stdin/stdout`, so I'm using IoImpl vs. the public interface. feels OK.

packages/cli/src/testHelpers/FakeIo.ts
```diff
@@ -1,19 +1,17 @@
...
-import type { Io } from '#src/IoImpl.ts';
-import { MemoryStream } from './index.ts';
+import { IoImpl } from '#src/IoImpl.ts';
+import { MemoryStream } from './MemoryStream.ts';
 
-export class FakeIo implements Io {
-  readonly stdout: MemoryStream;
-  readonly stderr: MemoryStream;
-  readonly stdin: PassThrough & { isTTY?: boolean };
+export class FakeIo extends IoImpl {
```
jcarver989: Seems odd for a fake to extend something called `IoImpl`.

packages/cli/src/IoImpl.ts
```diff
@@ -1,24 +1,58 @@
...
 
-export type Io = Public<IoImpl>;
+export interface Io {
+  readonly stdinIsTTY?: boolean;
+  readStdin(): Promise<string>;
+  writeStdout(chunk: string): void;
+  writeStderr(chunk: string): void;
+}
+
+type IoStreams = {
```
jcarver989: Would it make more sense to have `Writer` and `Reader` `extend` `Writeable` and `Readable` so you don't have to do intersection types here and could use an `interface` for consistency?

jcarver989 (approved):

packages/cli/src/testHelpers/FakeIo.ts
blimmer: Agree - fixed by making the public Io interface wider.

packages/cli/src/IoImpl.ts
blimmer: Agree - thanks

PR #92, opened 2026-05-12
fix: let plain link clicks navigate instead of opening the annotation popover

### LINKED ISSUE #90: Clicking links opens comment box instead
In the issue view, clicking a link opens the comment box instead of following the link.

Expected behavior may need a clearer interaction pattern, such as separate actions for adding a comment vs. opening a link.

Current direction: disable commenting on links by default so clicking a link opens the link. To comment on linked text, require dragging across the link to select/highlight it.

blimmer:
## Summary

In the annotation UI, clicking a rendered markdown link opened the comment popover instead of following the link, because every annotatable element — links included — was being intercepted by the activation hook. This fix adds a single bail-out in `useTargetActivation.activate()`: when the click/keypress source is inside an `<a href>` and there is no active text selection, return early so the browser runs its native navigation. Drag-selecting across link text still opens the annotation popover via the existing `mouseUp` selection path, and clicking a link that already carries an annotation still opens the editor (via the unrelated `handleClickCapture` in `useAnnotationInteractions`).

Closes #90.

## Review focus

- **The bail-out predicate.** `shouldDeferToDefaultAction` is structured as positive composition (`isLink(source) && hasCollapsedSelection()`) so a future GFM checkbox / `<details>` case can slot in as a sibling predicate. Today it only matches `a[href]`.
- **Why a selection check at all.** Browsers don't fire `click` after a drag with different mousedown/mouseup targets, so in practice the collapsed-selection guard isn't strictly necessary for the drag-to-annotate path — but it's the honest expression of intent ("step aside only when the user appears to want navigation, not annotation") and cheap.
- **Test coverage gap exposed.** The new "click on an already-annotated link opens the editor" test is the first case to exercise `useAnnotationInteractions.handleClickCapture` end-to-end. It needed `fireEvent.click` with explicit bounding-rect coords rather than `user.click(link)`, because `user.click` wasn't pinning `caretPositionFromPoint` into the link's text node in the headless vitest/Chromium environment. Flagged here in case anyone has a cleaner way.
- **Keyboard accessibility.** The activate path is shared with `handleKeyDown` (Enter/Space), so a focused link's Enter key now navigates rather than annotating — matching the native `<a href>` behavior. This removes the only keyboard-driven way to annotate a bare link; the design direction in the issue accepts that ("drag across the link to annotate").
- **Manual test fixture.** Adds a single markdown link to `samplePlans.ts` (Goals section, "Zod") so the `Plan/App > Default` Storybook story exercises the three scenarios end-to-end: click navigates, drag-select annotates, click on the now-annotated link reopens the editor.

## Commits

- [\`4bb4ce2\`](https://github.com/contextbridge/planbridge/commit/4bb4ce2b08bc2ecb4cec1c8d609e297e8b98e9ab) — fix: let plain link clicks navigate instead of opening the annotation popover
- [\`56fdc35\`](https://github.com/contextbridge/planbridge/commit/56fdc35535e65bf609dad2a1d4f04a4f115279d0) — chore(annotation): add a markdown link to samplePlan for manual testing

### DIFF
--- a/packages/annotation/src/App.test.tsx
+++ b/packages/annotation/src/App.test.tsx
@@ -282,6 +282,88 @@ describe('App', () => {
     expect(link).toHaveAttribute('rel', 'noreferrer');
   });
 
+  it('lets a plain click on a link navigate without opening the annotation popover', async () => {
+    const user = userEvent.setup();
+    renderApp({
+      initialPayload: { contentKind: 'plan', content: 'Check [the docs](https://example.com) for details.' },
+    });
+
+    const link = await screen.findByRole('link', { name: 'the docs' });
+    await waitFor(() => {
+      expect(link).toHaveAttribute('data-target-id');
+    });
+
+    let defaultPrevented: boolean | undefined;
+    link.addEventListener(
+      'click',
+      (event) => {
+        defaultPrevented = event.defaultPrevented;
+        event.preventDefault();
+      },
+      { once: true },
+    );
+
+    await user.click(link);
+
+    expect(defaultPrevented).toBe(false);
+    expect(screen.queryByTestId(annotationPopoverTestIds.container)).not.toBeInTheDocument();
+  });
+
+  it('opens the annotation popover when text inside a link is drag-selected', async () => {
+    renderApp({
+      initialPayload: { contentKind: 'plan', content: 'Check [the docs](https://example.com) for details.' },
+    });
+
+    const link = await screen.findByRole('link', { name: 'the docs' });
+    const text = link.firstChild as Text;
+
+    const range = document.createRange();
+    range.setStart(text, 0);
+    range.setEnd(text, text.length);
+    const selection = window.getSelection()!;
+    selection.removeAllRanges();
+    selection.addRange(range);
+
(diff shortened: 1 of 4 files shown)


blimmer:
@floodfx please confirm that this behavior is intuitive to you - see https://69fb70c537ec652eeadc0648-psifehziew.chromatic.com/?path=/story/plan-app--default (there's a link in the storybook now for you to test with)

packages/annotation/src/App.test.tsx
```diff
@@ -282,6 +282,88 @@ describe('App', () => {
...
+    const text = link.firstChild as Text;
+
+    const range = document.createRange();
+    range.setStart(text, 0);
+    range.setEnd(text, text.length);
+    const selection = window.getSelection()!;
+    selection.removeAllRanges();
+    selection.addRange(range);
+
+    fireEvent.mouseUp(screen.getByTestId(annotatedMarkdownTestIds.container));
```
jcarver989: would be cool to have a `drag({ target: text, from: 0, to: text.length })` helper.

jcarver989 (approved):

PR #94, opened 2026-05-12
refactor(cli): make handlers ResultAsync<void, AbortError> end-to-end

### LINKED ISSUE #86: refactor(cli): make handlers `ResultAsync<void, AbortError>` end-to-end (neverthrow)
### What's the problem?

The CLI's failure-handling style is split between two idioms today:

- `commands/hookCodex.ts` and `updater/UpdaterImpl.ts` use neverthrow (`ResultAsync.fromPromise`, `Result.fromThrowable`, `.match`, `.andThen`) — errors flow as values until the very end.
- Every other handler (`commands/plan.ts`, `commands/open.ts`, `commands/hookClaude.ts`, `commands/installStatus.ts`, `commands/update.ts`, `commands/harnessOperation.ts`, `harnesses/{Claude,Codex}Installer.ts`) uses `try`/`catch` + the shared `abort(ctx, command, kind, message)` helper that fuses log-level routing and `throw new CommanderError(...)` into one call.

The fused `abort()` works fine today. The reviewer's instinct on PR #83 ([discussion r3229535900](https://github.com/contextbridge/planbridge/pull/83#discussion_r3229535900)) was that it composes awkwardly with neverthrow because the error is never a value — every call site that wants to surface a failure has to either `throw` immediately or wrap a `.match` fail-arm around an `abort()` side effect. Today that's fine because only `hookCodex.ts` and the updater compose this way; as more handlers move to Result-bearing internals, the asymmetry compounds.

Direction-setting decision captured in the PR thread: long-term we want neverthrow to be the dominant control-flow style for the cli package.

### Proposed solution

Refactor handlers to return `ResultAsync<TResponse, AbortError>` end-to-end. Move the log + throw from `abort()` to a single boundary in `cli.ts` / `commands/index.ts`.

**1. Introduce a typed error class.**

\`\`\`ts
// packages/cli/src/commands/abort.ts (or a new errors.ts)
export class AbortError extends CommanderError {
  readonly level: 'warn' | 'error';
  readonly kind: 'input' | 'runtime' | 'cancelled';

  constructor(command: string, kind: 'input' | 'runtime' | 'cancelled', message: string) {
    const exitCode = kind === 'cancelled' ? 130 : 1;
    super(exitCode, `contextbridge.${command}.${kind}Error`, message);
    this.kind = kind;
    this.level = kind === 'runtime' ? 'error' : 'warn';
  }

  static input(command: string, message: string)    { return new AbortError(command, 'input', message); }
  static runtime(command: string, message: string)  { return new AbortError(command, 'runtime', message); }
  static cancelled(command: string, message: string) { return new AbortError(command, 'cancelled', message); }
}
\`\`\`

(The `'cancelled'` kind absorbs the ad-hoc `throw new CommanderError(130, ...)` in `commands/plan.ts:51`, `commands/open.ts:58`, and `prompter.ts:48`/`:67`/`:76`.)

**2. Convert handler signatures incrementally.**

\`\`\`ts
// before
export async function runOpen(ctx: CliContext, args: OpenArgs, deps?: AnnotationDependencies): Promise<void> { /* try/catch + abort */ }

// after
export function runOpen(ctx: CliContext, args: OpenArgs, deps?: AnnotationDependencies): ResultAsync<void, AbortError> { /* .andThen chain */ }
\`\`\`

Internal `try/catch` blocks around `Bun.file(...).text()`, schema parsing, and `runAnnotation` become `ResultAsync.fromPromise(...).mapErr(toAbort)` chains. The pattern is already established in `commands/hookCodex.ts:71-84` (`safeJsonParse → .andThen → .match`).

**3. Centralize the log + throw at the commander boundary.**

\`\`\`ts
// commands/index.ts
.action(async (path) => {
  const result = await runOpen(ctx, { path });
  result.match(
    () => undefined,
    (err) => {
      ctx.logger[err.level](err.message);
      throw err;
    },
  );
});
\`\`\`

One place handles log-level routing and the throw; impossible to forget either.

**4. Handler inventory and rough sequencing.**

| Handler | LOC-ish | Notes |
|---|---|---|
| `commands/plan.ts` | small | Mirror of `open.ts`; do them together. |
| `commands/open.ts` | small | New on this branch; freshest test coverage. |
| `commands/hookClaude.ts` | medium | Has the largest local `abort()` (different contract — always logs at error). Worth folding into `AbortError` with `kind: 'runtime'`. |
| `commands/hookCodex.ts` | medium | Partially done; finish wiring the boundary log+throw. |
| `commands/update.ts` | medium | |
| `commands/installStatus.ts` | small | |
| `commands/harnessOperation.ts` | medium | Iterates over installers — collect per-installer Results into a single combined failure. |
| `harnesses/{Claude,Codex}Installer.ts` | medium | Multiple `throw new CommanderError` sites; convert install/uninstall return types to Result. |

Suggested order: `plan` + `open` first (smallest, freshly tested), then hook handlers, then installers + harnessOperation, then `update`.

**5. Test-side migration.**

Every `expect(runX(...)).rejects.toBeInstanceOf(CommanderError)` site becomes `expect((await runX(...))._unsafeUnwrapErr()).toBeInstanceOf(AbortError)` (or a `.match` assertion). Doable per-handler in the same commit as the handler refactor; keep `just verify` green at each step.

- **Status quo** (Option A from the PR thread). Costs nothing today, scales poorly as more handler internals adopt Result-bearing IO wrappers.
- **Typed error + per-call-site log+throw** (Option B). Splits responsibility — every call site has to remember to log. Adds boilerplate without buying meaningful new capability.
- **Hybrid: typed `AbortError` + `fail(ctx, err)` helper that logs+throws** (Option D). Cheapest path; leaves the door open to end-to-end without forcing it. Worth considering as an interim if the full refactor is too big to land at once — we could land step 1 (the typed error + `fail()` helper) as a single commit, then convert handlers opportunistically.

### Additional context

- Original PR thread: [#83 r3229535900](https://github.com/contextbridge/planbridge/pull/83#discussion_r3229535900) (reviewer's nudge) and the deeper option breakdown discussed offline.
- Related follow-up: #84 (template terminology) and #85 (narrowing the `Io` interface). Independent of this work but adjacent — same PR, same handler files.
- The team neverthrow conventions are codified at `.contextbridge/rules/error-handling-neverthrow.md` — this refactor brings the cli package in line with the same conventions already applied to IO wrappers.

blimmer:
## Summary

Closes #86. Converts every CLI subcommand handler, installer, and the prompter to return `ResultAsync<T, AbortError>` end-to-end instead of the previous `try`/`catch` + fused `abort()` style. Introduces a typed `AbortError` (subclass of `CommanderError`) and centralizes the log + throw at the commander boundary via `handleCommandResult` / `throwLoggedAbort`, so the cli package now matches the team neverthrow conventions already applied to IO wrappers.

## Review focus

- **`commands/abort.ts`** — the new `AbortError` class, the `handleCommandResult` boundary helper, and the `toAbortError` / `abortable` adapters. Everything else hangs off this shape.
- **`cancelled` kind absorbs ad-hoc `CommanderError(130, ...)` throws** previously scattered across `plan.ts`, `open.ts`, and `prompter.ts`. Worth confirming the new error code strings still match what consumers (and the Sentry pinoIntegration filter) expect.
- **`harnessOperation.ts` + the installers** — multi-installer flow now collects per-installer Results into a single combined failure. Sanity-check the aggregation behavior.
- **Test migration** — `testHelpers/resultAssertions.ts` adds `expectOk` / `expectErr` helpers; the previous `rejects.toBeInstanceOf(CommanderError)` assertions are gone. New rule in `.contextbridge/rules/testing-patterns.md` documents the assertion style.

## Commits

- [`f52fd7f`](https://github.com/contextbridge/planbridge/commit/f52fd7ff63f100f0c6480fffaa0e6456730a33b8) — refactor(cli): make handlers ResultAsync<void, AbortError> end-to-end

### DIFF
--- a/.contextbridge/rules/testing-patterns.md
+++ b/.contextbridge/rules/testing-patterns.md
@@ -45,3 +45,5 @@ globs: ["packages/**/*.test.ts", "packages/**/*.test.tsx", "packages/**/*TestIds
   import { annotationPopoverTestIds } from './AnnotationPopover.tsx';
   screen.getByTestId(annotationPopoverTestIds.textarea);
   ```
+
+- **Use shared Result assertion helpers for neverthrow handlers.** When testing code that returns `Result` or `ResultAsync`, import `expectOk`, `expectOkValue`, or `expectErr` from the package's test helpers instead of using `.rejects` or repeating local `isOk()` / `isErr()` helpers. This keeps neverthrow tests explicit and consistent.
--- a/packages/cli/src/cli.test.ts
+++ b/packages/cli/src/cli.test.ts
@@ -6,7 +6,13 @@ import { describe, expect, it } from 'bun:test';
 import { Command } from 'commander';
 import { CLAUDE_MARKETPLACE_NAME, CLAUDE_PLUGIN_ID } from '#src/installers/ClaudeInstaller.ts';
 import { environment } from '#src/testFactories.ts';
-import { createStubContext, primeClaudeShellouts, readErrorLogs, stubClaudeState } from '#src/testHelpers/index.ts';
+import {
+  createStubContext,
+  primeClaudeShellouts,
+  readErrorLogs,
+  readWarnLogs,
+  stubClaudeState,
+} from '#src/testHelpers/index.ts';
 import { resolveCbCommand, runCli } from './cli.ts';
 
 const CLAUDE_BINARY = getHarness('claude').binaryName;
@@ -75,7 +81,8 @@ describe('runCli', () => {
 
     expect(exitCode).toBe(1);
     expect(io.stdout.text()).toBe('');
-    expect(readErrorLogs(logs).some((r) => r.msg.includes('unsupported hook_event_name: PreToolUse'))).toBe(true);
+    expect(readWarnLogs(logs).some((r) => r.msg.includes('unsupported hook_event_name: PreToolUse'))).toBe(true);
+    expect(readErrorLogs(logs)).toEqual([]);
   });
 
   it('surfaces unknown flags on hook claude as a CommanderError', async () => {
--- a/packages/cli/src/commands/abort.ts
+++ b/packages/cli/src/commands/abort.ts
@@ -1,18 +1,74 @@
+import { getErrorMessage } from '@contextbridge/shared/errors';
 import { CommanderError } from 'commander';
+import { ResultAsync } from 'neverthrow';
 import type { CliContext } from '#src/context.ts';
 
-/**
- * Log and throw a CommanderError for a user-recoverable ('input') or runtime
- * failure inside a subcommand handler.
- */
-export function abort(ctx: CliContext, command: string, kind: 'input' | 'runtime', message: string): never {
-  const { logger } = ctx;
-  // 'input' is user-recoverable — logged at warn so Sentry's pinoIntegration
-  // (error/fatal only) doesn't forward it. 'runtime' is a genuine failure.
-  if (kind === 'input') {
-    logger.warn(message);
-  } else {
-    logger.error(message);
+export type AbortKind = 'input' | 'runtime' | 'cancelled';
+export type AbortLevel = 'warn' | 'error';
+
+interface AbortErrorOptions {
+  readonly code?: string;
+  readonly exitCode?: number;
+}
+
+export class AbortError extends CommanderError {
+  readonly kind: AbortKind;
+  readonly level: AbortLevel;
+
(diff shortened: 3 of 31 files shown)


blimmer:
This is a huge change and doesn't feel that important to resolve right now. cc @jcarver989 - we can handle this some other time.

jcarver989:
> This is a huge change and doesn't feel that important to resolve right now. cc @jcarver989 - we can handle this some other time.

Yea I'm fine punting on this right now.

PR #95, opened 2026-05-12
feat: allow configuring PlanBridge port

floodfx:
## Summary
- add CONTEXTBRIDGE_PORT for hook-driven plan review port configuration
- add contextbridge plan --port for manual one-off runs
- preserve default OS-assigned port behavior via port 0 and document fixed-port usage with exe.dev as an example

## Test Plan
- just verify


README.md
```diff
@@ -59,6 +59,21 @@ Your plan content stays on your machine. No remote backend, no account, no API k
 - [Usage: Other agents](https://plan.contextbridge.ai/usage/other-agents/)
 - [Privacy & Telemetry](https://plan.contextbridge.ai/privacy/)
 
+## Port configuration
```
blimmer: We should make this change in the website instead, I think.

It'd be nice if we could auto-gen from source all of the environment config properties (not a requirement for this PR). However, I'd like to keep the README succinct.

Even more reason for me to do https://github.com/contextbridge/planbridge/issues/99 soon!

----

For now, could you:

1. Remove this config from here.
2. Create a PR in the private repo backing `tools/` to add this in the relevant place.
3. If you don't tackle auto-generating the `CONTEXTBRIDGE_*` env vars, could you open a github issue so we don't forget about it in the future?

packages/cli/src/commands/plan.ts
```diff
@@ -61,7 +66,21 @@ export function registerPlan(ctx: CliContext, program: Command): void {
...
     .argument('[path]', 'path to a file containing the plan (alternative to stdin)')
-    .action(async (path: string | undefined) => {
-      await runPlan(ctx, { path });
+    .option('--port <number>', 'serve the plan review browser UI on a specific port', parsePortOption)
+    .action(async (path: string | undefined, opts: { port?: number }) => {
+      await runPlan(ctx, { path, port: opts.port });
     });
 }
+
+function parsePortOption(value: string): number {
```
blimmer, line 75: This is duplicating logic that zod already has encoded. Can we lean on that instead?

blimmer (approved):
Only (soft) blocker for me is the README / website updates.

floodfx:
Addressed the review feedback:

- removed the fixed-port docs from the public README
- moved the docs to the website in contextbridge/planbridge-private#17
- opened follow-up issue #118 for auto-generated environment variable docs
- changed the manual `--port` parser to reuse the shared Zod-backed port parser from `environment.ts`

Verification:

- `just verify`
- `bun run --cwd tools/website test`
- `bun run --cwd tools format:check`

blimmer (approved):

PR #96, opened 2026-05-13
feat(storage): Add SQLite DB for local-only storage to support future features like plan diff and  git diff. 

jcarver989:
## Summary

This PR sets up the scaffolding required for a local-only sqlite DB + Drizzle ORM in a new `storage` package. The impetus is we have some upcoming features for plan-diffs and git-diff review that will require a data storage layer on the user's device. 

## Changes 

- Adds Drizzle and uses Bun's sqlite driver 
- Creates initial DB table +migration as an example
- Scaffolds code to auto-apply migrations on startup. 
- Checks in generated migration + snapshot files (for now).


packages/storage/scripts/generateMigrations.ts
```diff
@@ -0,0 +1,46 @@
+import { readdirSync, writeFileSync } from 'node:fs';
```
jcarver989: We generate a "index" of all migrations into a `migrations.ts` file so we have it available at runtime to hand to Drizzle's migration runner.

packages/storage/generated/drizzle/20260512235208_initial_schema/snapshot.json
```diff
@@ -0,0 +1,138 @@
+{
```
blimmer, line 1: It'd be nice to ignore this in a `.gitattributes` file, so they're collapsed in the PR review UI by default.

packages/storage/src/paths.ts
```diff
@@ -0,0 +1,43 @@
+import { chmodSync, mkdirSync } from 'node:fs';
+import { homedir as defaultHomedir } from 'node:os';
+import { dirname, isAbsolute, join } from 'node:path';
+import { Result } from 'neverthrow';
+import { StorageError, toStorageError } from './storageError.ts';
+
+export interface ResolveStoragePathOptions {
+  readonly env: Partial<Record<'CONTEXTBRIDGE_DB_PATH' | 'XDG_DATA_HOME' | 'HOME', string | undefined>>;
```
blimmer: Should this take a CliContext instead? Since we have a typed environment on that already?

packages/storage/scripts/generateMigrations.ts
blimmer: I think we can replace this codegen script with a bun macro. See https://github.com/contextbridge/planbridge/pull/105 for my suggestion.

packages/storage/src/db/schema/projects.ts
```diff
@@ -0,0 +1,19 @@
+import { index, sqliteTable, text, uniqueIndex } from 'drizzle-orm/sqlite-core';
+import { timestamps } from './timestamps.ts';
+
+export const PROJECT_VCS_KIND_VALUES = ['none', 'git'] as const;
+
+export const projects = sqliteTable(
+  'projects',
+  {
+    id: text('id').primaryKey(),
```
blimmer: Should this default to a new UUID?

packages/storage/src/db/schema/projects.ts
```diff
@@ -0,0 +1,19 @@
...
+    path: text('path').notNull(),
+    displayName: text('display_name').notNull(),
+    vcsKind: text('vcs_kind', { enum: PROJECT_VCS_KIND_VALUES }).notNull(),
+    vcsRootPath: text('vcs_root_path'),
```
blimmer, line 16: Should we have a `check` that this isn't set if vcs is none?

packages/storage/src/db/schema/timestamps.ts
```diff
@@ -0,0 +1,6 @@
+import { text } from 'drizzle-orm/sqlite-core';
+
+export const timestamps = {
+  createdAt: text('created_at').notNull(),
+  updatedAt: text('updated_at').notNull(),
```
blimmer: I think we want an `onUpdate` to set these automatically:

```ts
import { timestamp } from 'drizzle-orm/pg-core';

export const timestamps = {
  createdAt: timestamp('created_at', { withTimezone: true }).notNull().defaultNow(),
  updatedAt: timestamp('updated_at', { withTimezone: true })
    .notNull()
    .defaultNow()
    .$onUpdateFn(() => new Date()),
};
```

packages/storage/src/db/schema/projects.ts
```diff
@@ -0,0 +1,19 @@
+import { index, sqliteTable, text, uniqueIndex } from 'drizzle-orm/sqlite-core';
+import { timestamps } from './timestamps.ts';
+
+export const PROJECT_VCS_KIND_VALUES = ['none', 'git'] as const;
+
+export const projects = sqliteTable(
+  'projects',
+  {
+    id: text('id').primaryKey(),
+    path: text('path').notNull(),
```
blimmer, line 13: We'll want to canonicalize the path somehow (avoiding `~`, `$HOME`, `../`, etc.) when we set this

blimmer (approved):
a few things I think we should address, but looking like a great starting point!

packages/storage/scripts/generateMigrations.ts
jcarver989: Nice, didn't know about macros.

packages/storage/src/db/schema/projects.ts
jcarver989: Yes

packages/storage/src/db/schema/projects.ts
jcarver989: Yea, I was imagining we'd always write an absolute path and never use relative. If a user moved directories, in theory the remote url would still be the same and we could group by that to link sessions across.

packages/storage/src/paths.ts
jcarver989: Yea good call

PR #98, opened 2026-05-13
ci: surface zizmor findings via code scanning and address findings

blimmer:
## Summary

Now that planbridge is public, GitHub code scanning is available at no cost. The zizmor workflow now uploads its SARIF to code scanning (findings land in the Security tab instead of an artifact download), and this branch also clears the existing findings so the dashboard starts clean. As cleanup, the broken `build-contextbridge-image.yml` workflow — the only thing referencing a private action — is removed, which lets the zizmor job drop its app-token step and run with the default `GITHUB_TOKEN`.

## Review focus

- **`secrets: inherit` removed from `test.yml` callers** (main, pr, release). Verified `test.yml` references no `${{ secrets.* }}`; `bootstrap` uses `${{ github.token }}` which is unaffected. If a future test step grows a secret dependency, this will need to be added back.
- **`release.yml` permissions narrowed.** `contents: write` and `id-token: write` moved from workflow level to the `release` job only; workflow level is now `contents: read`. The `release` job still has `packages: read` for the bootstrap action's GitHub Packages auth.
- **`persist-credentials: false` added to checkouts in `release.yml`** (which uses `token: ${{ steps.app-token.outputs.token }}`). GoReleaser uses env-var tokens (`GITHUB_TOKEN`, `HOMEBREW_TAP_GITHUB_TOKEN`), not git-config credentials, so dropping git-config persistence is safe — but worth a second look in case any future step relies on `git push` from the checkout.
- **`dependabot-cooldown` suppressed, not fixed.** zizmor wants `default-days >= 7`; kept at 3 with `# zizmor: ignore[dependabot-cooldown]` — security advisories bypass cooldown anyway, and 3 days is a deliberate tradeoff for faster bumps.
- **`auto-approve.yml` left untouched.** Its `dangerous-triggers` finding (uses `pull_request_target`) is the one remaining issue; the file will be removed in a follow-up.

## Commits

- [`c6b2c40`](https://github.com/contextbridge/planbridge/pull/98/commits/c6b2c40729e89ed237e124602794f2aed46aa4bf) — swap zizmor SARIF artifact upload for `github/codeql-action/upload-sarif`; add `security-events: write`.
- [`1b2b469`](https://github.com/contextbridge/planbridge/pull/98/commits/1b2b46922cfe77e60c6e4819d7e53ef0219d5141) — delete the broken `build-contextbridge-image.yml` (the only private-action reference) and drop the now-unnecessary app-token step from the zizmor workflow.
- [`db9a776`](https://github.com/contextbridge/planbridge/pull/98/commits/db9a77653d83121ed478e775385f8b468b9d7de4) — address remaining zizmor findings: pin 14 actions to SHA, fix 3 template-injections via env-var indirection, add `persist-credentials: false` to 7 checkouts, narrow `release.yml` permissions, drop `secrets: inherit` from `test.yml` callers.


.github/dependabot.yml
```diff
@@ -16,7 +16,7 @@ updates:
     schedule:
       interval: "weekly"
     cooldown:
-      default-days: 3
+      default-days: 3 # zizmor: ignore[dependabot-cooldown]
```
blimmer, line 19: zizmor recommends 7 days, but I feel like 3 is fine. Happy to bump up if you feel differently.

.github/workflows/zizmor.yml
```diff
@@ -41,12 +31,10 @@ jobs:
       - name: Run zizmor
         run: uvx zizmor --format=sarif .github/ > zizmor.sarif
         env:
-          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Upload zizmor SARIF
-        if: always()
-        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+      - name: Upload zizmor SARIF to code scanning
```
blimmer, line 36: Instead of just having a report file, this will now show issues in the Security tab.

.github/workflows/zizmor.yml
```diff
@@ -18,18 +18,8 @@ jobs:
     permissions:
       contents: read
       actions: read # zizmor reads workflow run metadata for online audits
+      security-events: write # upload SARIF to GitHub code scanning
     steps:
-      - name: Generate ContextBridge app token
```
blimmer, line 22: Since we removed the private build-contextbridge-image.yml workflow, we don't need to use the app token (which feels safer).

.github/workflows/pr.yml
```diff
@@ -38,4 +40,3 @@ jobs:
     permissions:
       contents: read
       packages: read
-    secrets: inherit
```
blimmer, line 41: This step doesn't need access to secrets.

.github/workflows/main.yml
```diff
@@ -40,4 +42,3 @@ jobs:
     permissions:
       contents: read
       packages: read
-    secrets: inherit
```
blimmer, line 43: This step doesn't need access to secrets

.github/workflows/build-contextbridge-image.yml
blimmer, line 1: Since this is failing anyway (and we want to move to pushing the image directly to ECR), I'm removing it to secure the zizmor check (no app token).

.github/dependabot.yml
jcarver989: Most of the major pkg attacks are resolved within 3 days so I think it's ok for now.

jcarver989 (approved):
Thanks!

PR #103, opened 2026-05-13
fix: expose PlanBridge open skill in Codex

blimmer:
## Summary

Fixes the Codex-installed PlanBridge open skill so Codex exposes it as `$planbridge-open` instead of `$open`. The installer continues to use the Claude skill asset as the source of truth and rewrites only the frontmatter name for Codex.

## Review focus

The key behavior is in the Codex skill bundling path: Claude namespaces plugin skills as `/planbridge:open`, while Codex displays the skill frontmatter name directly. The test coverage should make sure we do not reintroduce a duplicated Codex skill file or lose the `planbridge-` prefix.

## Commits

- [`61a0743`](https://github.com/contextbridge/planbridge/commit/61a07432e93890089cfcbb5e3e6ecad547e1292d) — derive the Codex skill name from the Claude skill source.

### DIFF
--- a/packages/cli/src/bundledSkills.ts
+++ b/packages/cli/src/bundledSkills.ts
@@ -1,9 +1,8 @@
 // Skill assets embedded into the CLI binary at build time. The Codex installer
 // writes each entry to disk under `~/.agents/skills/planbridge-<name>/SKILL.md`;
-// the Claude side ships the same files as plugin assets at
+// the Claude side ships plugin assets at
 // `harnessIntegrations/claude/skills/<name>/SKILL.md`. A sync test
-// (`bundledSkills.test.ts`) asserts each embedded copy stays in lockstep with
-// its on-disk Claude source.
+// (`bundledSkills.test.ts`) asserts Codex bodies only rewrite the skill name.
 //
 // To add a new skill: drop `harnessIntegrations/claude/skills/<name>/SKILL.md`
 // in place, then add a single line to the `bundledSkills` array below — the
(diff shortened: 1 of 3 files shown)


packages/cli/src/bundledSkills.ts
```diff
@@ -1,20 +1,27 @@
...
 }
 
-export const bundledSkills: readonly BundledSkill[] = [{ name: 'open', body: openSkill }];
+export const bundledSkills: readonly BundledSkill[] = [{ name: 'open', body: toCodexSkill('open', openSkill) }];
+
+/**
+ * Claude displays plugin skills with the plugin namespace (`/planbridge:open`),
+ * but Codex displays the skill's frontmatter name directly (`$planbridge-open`).
+ */
+function toCodexSkill(name: string, body: string): string {
```
blimmer, line 25: Do we hate this? Should we come up with some other way of having the canonical skill body, then generating the claude vs. codex version?

jcarver989: I think what's a bit awkward about this is we're treating `harnessIntegrations/claude/skills/...` as the source of truth, rather than storing something like a `SkillDefinition` type/md file that acts as a harness agnostic source of truth and renders to a harness specific skill format + location -- e.g. Claude, Codex, ...etc compatible .

jcarver989 (approved):

packages/cli/src/bundledSkills.ts
blimmer: Yeah, I was being lazy. I didn't really want to introduce a whole codegen step, but I suppose we could.

Do you think we should do that now, since this is our first skill?

jcarver989: I think it's worth doing, but don't necessarily have to do it in this PR. Seems like we could probably store the skills in the generic spec [here](https://agentskills.io/specification) and just map the frontmatter to the format the harness wants.

blimmer: Sounds good. We're kinda blocked right now because zizmor nerfed our GITHUB_TOKEN, so I might take a look now.

blimmer:
I'm going to take the new approach we discussed (with codegen) in a new PR.

blimmer:
Actually, since Josh is about to be AFK and my skill codegen PR isn't ready yet, I'm going to do this for v0.5.0 and the ShowHN thread.

PR #104, opened 2026-05-13
fix: stop disabling deprecated codex_hooks feature flag

### LINKED ISSUE #100: bug: disabling codex_hooks via `codex features disable` still triggers Codex deprecation warning
### What happened?

When PlanBridge installs the Codex hook, it migrates the deprecated `codex_hooks` feature flag off by running `codex features disable codex_hooks`. That command sets `[features].codex_hooks = false` in `~/.codex/config.toml` rather than removing the key. Codex's deprecation warning fires whenever the key is **present at all** (regardless of `true`/`false`), so users still see the warning every time they start Codex after running `contextbridge install` (or fresh `status`/upgrade flows that re-run the migration).

The fix is to stop relying on `codex features disable codex_hooks` and instead directly remove the `codex_hooks` key from `[features]` in `config.toml` using a TOML-aware patcher (e.g. `toml-patch`), so the key is deleted rather than set to `false`.

Relevant code: `packages/cli/src/harnesses/CodexInstaller.ts:332-335` (`enableCodexHookFeatureFlags`).

### contextbridge version

0.5.0-alpha.2

### Operating system

macOS

### Steps to reproduce

1. Have Codex CLI >= 0.129.0 installed with a legacy `[features].codex_hooks = true` entry in `~/.codex/config.toml` (or simply run `codex features enable codex_hooks` first to seed it).
2. Run `contextbridge install` (or any flow that invokes `enableCodexHookFeatureFlags`) and let it complete successfully.
3. Inspect `~/.codex/config.toml` — `[features].codex_hooks` is still present, now set to `false`.
4. Start Codex (`codex`). Observe the deprecation warning printed on startup.

After PlanBridge runs its feature-flag migration, the legacy `codex_hooks` key should be removed entirely from `~/.codex/config.toml` so Codex no longer emits the deprecation warning on startup.

### What actually happened?

`codex features disable codex_hooks` writes `codex_hooks = false` into the `[features]` table instead of removing the key. Codex still detects the presence of the deprecated flag and prints:

```shell
⚠ `[features].codex_hooks` is deprecated. Use `[features].hooks` instead.
Enable it with `--enable hooks` or `[features].hooks` in config.toml. See
https://developers.openai.com/codex/config-basic#feature-flags for details.
```

### Relevant logs or stack trace

```shell
⚠ `[features].codex_hooks` is deprecated. Use `[features].hooks` instead.
Enable it with `--enable hooks` or `[features].hooks` in config.toml. See
https://developers.openai.com/codex/config-basic#feature-flags for details.
```

### Anything else

- Suggested approach: replace the `codex features disable codex_hooks` call in `enableCodexHookFeatureFlags` (`packages/cli/src/harnesses/CodexInstaller.ts:334`) with a direct edit of `~/.codex/config.toml` using `toml-patch` (or similar) that removes the `codex_hooks` key from the `[features]` table, leaving the rest of the file's formatting/comments intact. If `[features]` becomes empty after the removal, decide whether to also drop the empty table.
- Tests to update: `packages/cli/src/harnesses/CodexInstaller.test.ts` currently asserts the `features disable codex_hooks` shell call; that should switch to asserting the resulting `config.toml` contents.

blimmer:
## Summary

`enableCodexHookFeatureFlags` was running `codex features disable codex_hooks` as a tidy-up step after enabling the new `hooks` flag, but that command writes `codex_hooks = false` into the user's `[features]` table rather than removing the key. Codex's deprecation warning fires whenever the key is present at all, so users kept seeing it on every Codex startup after `contextbridge install`.

Drop the call entirely. Fresh installs on Codex `0.129.0+` no longer touch the deprecated key, so it stays absent and the warning never fires. Legacy installs that already wrote `codex_hooks` to the user's config will need to hand-remove the key from `~/.codex/config.toml`; that path is documented separately in the Codex usage docs (private repo PR).

Closes #100.

## Review focus

- The alternative — direct TOML edits via `toml-patch` — was considered and rejected. `toml-patch` is unmaintained (last release 2019), and surgically rewriting the user's `config.toml` for cosmetic cleanup introduces failure modes (malformed TOML, lost comments) that aren't worth the trade. Codex's deprecation warning has a bounded lifespan and is self-documenting, so leaving the migration to users is reasonable.
- `tryRunCodexCommand` was deleted as part of this change — it was only used for the now-gone disable call.
- `cli.test.ts` still has a test (`'routes install codex without parsing Codex config.toml'`) that seeds a malformed `config.toml` and asserts install succeeds. That test now serves double duty as a regression guard against the rejected alternative.

## Commits

- [`8a60e28`](https://github.com/contextbridge/planbridge/commit/8a60e28) — fix: stop disabling deprecated codex_hooks feature flag

### DIFF
--- a/packages/cli/src/harnesses/CodexInstaller.ts
+++ b/packages/cli/src/harnesses/CodexInstaller.ts
@@ -76,7 +76,7 @@ export class CodexInstaller extends ScopedHarnessInstaller {
     const hooksSource = await readOptionalText(hooksPath);
     const nextHooks = upsertPlanBridgeHookJson(hooksSource);
     await writeFile(hooksPath, nextHooks);
-    await enableCodexHookFeatureFlags(ctx, this.descriptor.binaryName);
+    await enableCodexHooksFeature(ctx, this.descriptor.binaryName);
     for (const skill of bundledSkills) {
       await writeSkill(ctx, skill);
     }
@@ -329,9 +329,8 @@ async function requireSupportedCodexVersion(ctx: CliContext, binaryName: string)
   }
 }
 
-async function enableCodexHookFeatureFlags(ctx: CliContext, binaryName: string): Promise<void> {
+async function enableCodexHooksFeature(ctx: CliContext, binaryName: string): Promise<void> {
   await runCodexCommand(ctx, binaryName, ['features', 'enable', 'hooks'], 'features enable hooks');
-  await tryRunCodexCommand(ctx, binaryName, ['features', 'disable', 'codex_hooks'], 'features disable codex_hooks');
 }
 
 async function runCodexCommand(
@@ -355,32 +354,6 @@ async function runCodexCommand(
   }
 
   return result;
-}
-
-async function tryRunCodexCommand(
-  ctx: CliContext,
-  binaryName: string,
-  args: readonly string[],
-  label: string,
-): Promise<void> {
-  const { commandRunner, logger } = ctx;
-  try {
-    const result = await commandRunner.run(binaryName, args);
-    if (result.exitCode === 0) {
-      logger.debug(
-        { args, exitCode: result.exitCode, stdout: result.stdout, stderr: result.stderr },
-        `${binaryName} ${label}`,
-      );
-      return;
-    }
-
-    logger.warn(
-      { args, exitCode: result.exitCode, stdout: result.stdout, stderr: result.stderr },
-      `${binaryName} ${label} failed; continuing`,
-    );
-  } catch (err) {
-    logger.warn({ err, args }, `${binaryName} ${label} failed; continuing`);
-  }
 }
 
 function parseCodexVersion(source: string): string | null {
--- a/packages/cli/src/cli.test.ts
+++ b/packages/cli/src/cli.test.ts
@@ -119,7 +119,6 @@ describe('runCli', () => {
     commandRunner.setWhich(CODEX_BINARY, '/usr/local/bin/codex');
     commandRunner.on(CODEX_BINARY, ['--version']).resolves({ stdout: 'codex-cli 0.129.0\n' });
     commandRunner.on(CODEX_BINARY, ['features', 'enable', 'hooks']).resolves();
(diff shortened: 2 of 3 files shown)


packages/cli/src/harnesses/CodexInstaller.ts
```diff
@@ -329,9 +329,8 @@ async function requireSupportedCodexVersion(ctx: CliContext, binaryName: string)
   }
 }
 
-async function enableCodexHookFeatureFlags(ctx: CliContext, binaryName: string): Promise<void> {
+async function enableCodexHooksFeature(ctx: CliContext, binaryName: string): Promise<void> {
   await runCodexCommand(ctx, binaryName, ['features', 'enable', 'hooks'], 'features enable hooks');
-  await tryRunCodexCommand(ctx, binaryName, ['features', 'disable', 'codex_hooks'], 'features disable codex_hooks');
```
blimmer, line 334: I originally had an agent test this and it said something like "oh, when it's set to `false`, it doesn't show the deprecation warning!" ... but it was wrong.

My thought is that we just don't mess with this and people can clean up their old flag. The benefit of this change is that we're not re-introducing the deprecated flag once it's already been cleaned up.

If you think we should be more aggressive, we can clean up the old flag for them (since we already gate on > 0.129.0 where they introduced the deprecation).

blimmer: See also https://github.com/contextbridge/planbridge-private/pull/16 for docs on this issue.

jcarver989: Yea re: testing , I think we'd ultimately want to run the agent in a dockerfile and capture the tty output and just write assertions on that for e2e tests (vs asking an agent, b/c agents will lie).

jcarver989 (approved):

packages/cli/src/harnesses/CodexInstaller.ts
blimmer: Agreed! (someday)

PR #105, opened 2026-05-13
refactor(storage): replace migration codegen with Bun macro

blimmer:
Suggestion for #96 — feel free to apply or discard.

## Why

`scripts/generateMigrations.ts` + the checked-in `generated/migrations.ts` are codegen plumbing that Bun lets us delete. A macro reads `generated/drizzle/**` at transpile time and inlines the journal directly, so:

- No second step in `db:generate`
- No checked-in `generated/migrations.ts` to drift out of sync
- `timestamp` now uses `drizzle-orm`'s own `formatToMillis` instead of a wrong `parseInt` of the directory prefix

## What's a Bun macro?

A regular function call whose *return value* is evaluated at transpile time and inlined into the AST as a literal:

```ts
import { loadStorageMigrations } from './loadMigrations.macro.ts' with { type: 'macro' };
export const storageMigrationsJournal = loadStorageMigrations();
```

Bun runs `loadStorageMigrations()` once while parsing this file (in `bun run`, `bun test`, and `bun build --compile`), then replaces the call with the resulting array of `{ name, timestamp, sql }` literals. The macro can use `node:fs`, `node:path`, etc.; the consumer sees only inlined data. Docs: https://bun.sh/docs/bundler/macros

## Caveats

- **Bun-only.** Vitest/esbuild won't evaluate the macro. Fine for `packages/storage` (already Bun-locked via `bun:sqlite`), but non-Bun consumers would break.
- The macro reads the FS at transpile time, so `generated/drizzle/**` still needs to exist on disk — which is already the contract (`db:generate` → check in).


jcarver989 (approved):
Cool , didn't know about bun macros, thanks!

PR #108, opened 2026-05-13
feat: canonical SKILL.md sources with per-harness codegen

### LINKED ISSUE #107: feat: canonical SKILL.md sources with per-harness codegen
### What's the problem?

`packages/cli/src/bundledSkills.ts` treats `harnessIntegrations/claude/skills/<name>/SKILL.md` as the authoritative source-of-truth and produces the Codex variant via a runtime `body.replace('name: <name>', 'name: planbridge-<name>')` (added in #103 to fix a Codex naming bug — see the [review thread](https://github.com/contextbridge/planbridge/pull/103#discussion_r3236034990)).

Two problems with that shape:

1. **Conceptually awkward.** The Claude file is named "claude" but doubles as the canonical source, which bakes one harness's conventions (plugin-namespaced `name`) into a position that should be harness-agnostic. A third harness with different conventions would need its own runtime rewriter.
2. **Fragile.** String `.replace()` on a frontmatter line works for one skill with one trivial divergence. Future skills will diverge in more places — different optional frontmatter fields, body prose that names the invocation prefix (`/planbridge:foo` vs `$planbridge-foo`), per-harness `compatibility` constraints, etc.

The forward look (per maintainer brainstorming): several more skills (e.g. `review`) plus likely a third harness (Cursor, Cline, Gemini CLI). Enough volume to justify a canonical source + renderer layer up front.

### Proposed solution

Introduce a `@contextbridge/skills` workspace package that owns the canonical SKILL.md sources (per the [agentskills.io specification](https://agentskills.io/specification)), per-harness renderer functions, and a drift-check script:

```
packages/skills/
├── sources/<name>/SKILL.md                       # hand-edited, agentskills.io frontmatter
├── src/
│   ├── parser.ts                                 # front-matter + zod-validated frontmatter
│   ├── render.ts                                 # render(skill, CLAUDE | CODEX) → string
│   ├── codex.ts                                  # bundledSkills export for the CLI binary
│   └── ...
└── scripts/
    ├── generate.ts                               # writes both committed outputs
    └── check.ts                                  # drift detection (wired into CI lint)
```

Rendered outputs are **committed** for review visibility and drift-checked in CI:

- `harnessIntegrations/claude/skills/<name>/SKILL.md` — read off disk by Claude's plugin marketplace.
- `harnessIntegrations/codex/skills/planbridge-<name>/SKILL.md` — embedded into the CLI binary via `with { type: 'text' }`, written to `~/.agents/skills/planbridge-<name>/SKILL.md` on Codex install.

Body content may include `{{invocation}}` and `{{harnessName}}` Handlebars placeholders that the renderer substitutes per harness. The renderer also rewrites the frontmatter `name` field for Codex (Claude's plugin manifest contributes the namespace; Codex displays the name directly).

`CodexInstaller` imports `bundledSkills` directly from `@contextbridge/skills/codex`, dropping the CLI-side `SKILL_ID_PREFIX` constant, `skillId(name)` helper, and `bundledSkills.ts` registry.

- **Top-level `skills/`** with renderer logic elsewhere — splits "source files" and "code that consumes them" across the tree without a clear win. Workspace package keeps the conceptual unit together.
- **`harnessIntegrations/skills/`** — overloads the `harnessIntegrations/` umbrella (which today holds outputs harnesses read verbatim) with sources, blurring the source/output distinction.
- **YAGNI on the codegen layer** — keep harden the string `.replace()`. Rejected because the next skill is imminent and the forward look points at multiple harnesses.

### Additional context

PR #103 review thread: https://github.com/contextbridge/planbridge/pull/103#discussion_r3236034990

blimmer:
## Summary

Introduce two new shared packages and migrate the CLI to consume them:

- **`@contextbridge/harness`** — a leaf, data-only package owning the canonical `HarnessDescriptor` shape and the registry of agent harnesses PlanBridge knows about (claude, codex, gemini, cursor, aider, opencode, aether). Today only claude and codex are installable + skill-renderable.
- **`@contextbridge/skills`** — canonical `sources/<name>/SKILL.md` per the [agentskills.io specification](https://agentskills.io/specification), with per-harness rendering driven by the descriptor's `skillRendering` rules. Generated outputs land at `harnessIntegrations/<harness-id>/skills/<name>/SKILL.md`. Drift detection (`skills:check`) runs in `just verify` and CI.

The CLI migrates to consume both packages, and `packages/cli/src/harnesses/` is renamed to `packages/cli/src/installers/` to resolve the naming collision — the CLI directory holds installer machinery (`ClaudeInstaller`/`CodexInstaller`, abstract bases) tightly coupled to `CliContext`, Commander, and `node:fs`; the new package holds pure descriptor data.

`CodexInstaller` imports `bundledSkills` directly from `@contextbridge/skills/codex` across the workspace boundary — `bun build --compile` follows the `with { type: 'text' }` import and bakes the SKILL.md content into the standalone binary.

Closes #107.

## Review focus

- **The `with { type: 'text' }` cross-workspace import in `packages/skills/src/codex.ts`** — verified empirically that `bun build --compile` follows it across packages and emits the body into the standalone binary even after the producer directory is moved away. If the bundler ever changes that behavior, the `Skills drift check` step in CI will still catch a stale committed Codex artifact, but the CLI binary would lose the embedded skill silently — worth a smoke test on the compiled binary if we're nervous.
- **`@contextbridge/harness` is intentionally a leaf with no workspace dependencies.** Descriptors hold data only; CLI/filesystem logic stays in `@contextbridge/cli`. Both `@contextbridge/skills` (data → rendered Markdown) and `@contextbridge/cli` (data → installer dispatch) consume the registry without dragging `CliContext`/Commander/`node:fs` into a shared package.
- **Frontmatter parser uses `front-matter` (not `gray-matter`)**: `gray-matter`'s cache uses `Object.assign({}, cached)`, which drops its non-enumerable `matter` property on the second call with identical input. That bites our test suite, where the same fixture appears in multiple cases. `front-matter` doesn't cache.
- **Byte-equality of generated harness artifacts** is preserved — `bun run skills:check` regenerates in-memory and fails on byte-diff against committed `harnessIntegrations/<id>/skills/...` files.
- **`BundledSkill.installId` is derived from each entry's frontmatter** in `packages/skills/src/codex.ts` — adding a future skill is one new import line plus one entry in the `bodies` array; no duplicated `'planbridge-<name>'` literal.

## Commits

- [`98a8789`](https://github.com/contextbridge/planbridge/commit/98a8789) — introduce `@contextbridge/harness` package (data-only descriptor registry).
- [`9058553`](https://github.com/contextbridge/planbridge/commit/9058553) — new `@contextbridge/skills` package with sources, parser, renderer, drift check, and CI wiring; committed harness outputs.
- [`bef7b11`](https://github.com/contextbridge/planbridge/commit/bef7b11) — CLI consumes both new packages; drops the local harness registry and the local `bundledSkills` module.
- [`c24d4b7`](https://github.com/contextbridge/planbridge/commit/c24d4b7) — rename `packages/cli/src/harnesses/` → `packages/cli/src/installers/` (resolves naming collision; reflects what the files actually are).
- [`5885d7e`](https://github.com/contextbridge/planbridge/commit/5885d7e) — rename rule file to `planbridge-skills.md` and refresh root `AGENTS.md` package tree.


packages/harness/src/index.ts
```diff
@@ -0,0 +1,14 @@
+export type {
```
blimmer, line 1: Now `packages/skills` and `packages/cli` need to know about harness details. That drives a need to hoist the harness registry details into a new package to avoid circular deps.

.github/workflows/test.yml
```diff
@@ -32,6 +32,8 @@ jobs:
         run: bun run lint
       - name: Typecheck
         run: bun run typecheck
+      - name: Skills drift check
+        run: bun run skills:check
```
blimmer, line 36: I want to actually have the harness skills in `src/harnessIntegrations` so people can easily spot-check what they do. This is also a requirement for Claude Code because they read the files from the `git clone`. Therefore, we can't use bun macros. 

Because of that, I need to make sure that the skills are up-to-date (someone didn't change them and forget to codegen)

packages/skills/src/codex.ts
```diff
@@ -0,0 +1,14 @@
...
+import { parseSkill } from './parser.ts';
+
+export interface BundledSkill {
+  /** On-disk install directory name, e.g. `planbridge-open`. Matches `~/.agents/skills/<installId>/`. */
+  readonly installId: string;
+  /** Pre-rendered Codex SKILL.md content embedded at build time. */
+  readonly body: string;
+}
+
+export const bundledSkills: readonly BundledSkill[] = [openSkill].map((body) => ({
```
blimmer, line 11: Since we're not (yet) using a codex plugin, I'm bundling the rendered skill content with the `bun` binary.

packages/skills/src/render.ts
```diff
@@ -0,0 +1,13 @@
...
+import type { CanonicalSkill } from './skillSchema.ts';
+
+export function render(skill: CanonicalSkill, harness: HarnessDescriptor): string {
+  const rules = harness.skillRendering;
+  if (!rules) {
+    throw new Error(`Harness ${harness.id} has no skill rendering rules`);
+  }
+  const frontmatter = { ...skill.frontmatter, name: rules.installName(skill.frontmatter.name) };
+  const frontmatterText = yamlStringify(frontmatter, { lineWidth: 0 });
+  return `---\n${frontmatterText}---\n\n${skill.body}`;
```
blimmer, line 12: Sadly, the frontmatter package only reads and doesn't serialize. Feels OK.

package.json
```diff
@@ -9,12 +9,14 @@
   "scripts": {
     "dev": "bun run --cwd packages/cli dev",
     "build": "just build",
-    "test": "bun run --cwd packages/shared test && bun run --cwd packages/context test && bun run --cwd packages/server test && bun run --cwd packages/storage test && bun run --cwd packages/ui test && bun run --cwd packages/harness test && bun run --cwd packages/cli test && bun run --cwd packages/annotation test",
+    "test": "bun run --cwd packages/shared test && bun run --cwd packages/context test && bun run --cwd packages/server test && bun run --cwd packages/storage test && bun run --cwd packages/ui test && bun run --cwd packages/harness test && bun run --cwd packages/skills test && bun run --cwd packages/cli test && bun run --cwd packages/annotation test",
```
blimmer, line 12: Man, I miss turborepo.

packages/cli/src/commands/install.test.ts
```diff
@@ -1,20 +1,20 @@
 import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs';
```
blimmer, line 1: This commit just rejiggers the cli package to consume the new harnesses/skills packages.

packages/cli/src/commands/harnessOperation.ts
```diff
@@ -1,9 +1,9 @@
 import { getErrorMessage } from '@contextbridge/shared/errors';
```
blimmer, line 1: This commit you can completely skim. All it does is change `packages/cli/src/harnesses` to `packages/cli/src/installers`. The existing naming was confusing because we now have a harnesses package.

packages/harness/src/registry.ts
```diff
@@ -0,0 +1,53 @@
...
+    skillRendering: {
+      installName: (name) => `planbridge-${name}`,
+      outDirFromRepoRoot: 'harnessIntegrations/codex/skills',
+    },
+  },
+  { id: 'gemini', displayName: 'Gemini CLI', binaryName: 'gemini' },
+  { id: 'cursor', displayName: 'Cursor', binaryName: 'cursor' },
+  { id: 'aider', displayName: 'Aider', binaryName: 'aider' },
+  { id: 'opencode', displayName: 'opencode', binaryName: 'opencode' },
+  { id: 'aether', displayName: 'Aether', binaryName: 'aether' },
```
jcarver989, line 28: Ha! Appreciate adding this

packages/harness/src/registry.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export const HARNESSES: readonly HarnessDescriptor[] = [
+  {
+    id: 'claude',
+    displayName: 'Claude Code',
+    binaryName: 'claude',
+    installUrl: 'https://docs.claude.com/en/docs/claude-code/setup',
+    skillRendering: {
+      installName: (name) => name,
+      outDirFromRepoRoot: 'harnessIntegrations/claude/skills',
```
jcarver989: Not a big deal but might consider `destinationDir` or `destDir`.

packages/skills/src/parser.test.ts
```diff
@@ -0,0 +1,63 @@
+import { describe, expect, it } from 'bun:test';
+import { parseSkill } from './parser.ts';
+
+describe('parseSkill', () => {
+  it('extracts frontmatter and body', () => {
+    const source = `---
+name: open
+description: Open a thing for review.
+---
```
jcarver989: would be nice to have a `frontmatter({ ... })` helper

packages/skills/src/skillSchema.ts
```diff
@@ -0,0 +1,23 @@
...
+    license: z.string().optional(),
+    compatibility: z.string().max(500).optional(),
+    metadata: z.record(z.string(), z.string()).optional(),
+    'allowed-tools': z.string().optional(),
+  })
+  .strict();
+
+export type SkillFrontmatter = z.infer<typeof SkillFrontmatterSchema>;
+
+export interface CanonicalSkill {
```
jcarver989: Would be tempted to just call this `Skill` since I think it'd be clear it's the canonical version we render from.

packages/skills/src/codex.ts
jcarver989, line 11: This looks like it'd require manual appending to this array every time we add a new skill? We might want to flip that such that the skill spec itself declares which harnesses we want to render to, and we could filter using that when rendering.

packages/skills/src/loadAll.ts
```diff
@@ -0,0 +1,21 @@
+import { readFileSync, readdirSync } from 'node:fs';
+import { join } from 'node:path';
+import { parseSkill } from './parser.ts';
+import type { CanonicalSkill } from './skillSchema.ts';
+
+export function loadAllFrom(sourcesDir: string): CanonicalSkill[] {
```
jcarver989: This PR has a lot of tiny files with 1-2 exported functions, I'd be tempted to co-locate the skill interface, parsing, and file/io to read/write them into a `skills.ts` file, but that's more out of personal preference.

jcarver989 (approved):

packages/skills/src/loadAll.ts
blimmer: Agreed - I think this is part of the problem, at least for the parts that deal with the schema. https://github.com/contextbridge/planbridge/blob/4e55c64dd5091ae190d158c6833796e049bd072f/AGENTS.md#L89

packages/skills/src/codex.ts
blimmer: The problem here is that we need to bundle the skills in the binary via distinct `import` statements, so I think this needs to stay here for now. We can always revisit this in the future, though

PR #115, opened 2026-05-14
feat: tell Codex to run contextbridge open outside the sandbox

### LINKED ISSUE #110: feat: tell Codex to run contextbridge open outside the sandbox
### What's the problem?

The Codex `planbridge-open` skill currently tells agents to run `contextbridge open <path>` like a normal command. In Codex, that usually starts the local annotation server, but the automatic browser launch depends on a GUI/desktop action and can fail or no-op from the default sandbox.

This makes `$planbridge-open /path/to/file.md` confusing: PlanBridge may print a localhost URL, but the browser does not reliably open automatically unless the command is run outside the sandbox.

### Proposed solution

Update the Codex-installed `planbridge-open` skill so it explicitly tells Codex to run `contextbridge open` with sandbox escalation, for example using `sandbox_permissions: "require_escalated"` and an appropriate justification. For path-based usage, the guidance should also suggest a reasonably scoped reusable approval such as `prefix_rule: ["contextbridge", "open"]`.

Implementation should avoid a one-off hardcoded appendix in the renderer. We will want:

- a way to add harness-specific details to a given skill, so Codex can receive operational guidance that Claude or other harnesses should not see
- a way to easily reuse standard Codex sandbox messaging, because future `contextbridge` commands will likely need to tell Codex to break out of the sandbox for browser, desktop, or other local integration behavior

A direct Codex-only text block hardcoded in the skill renderer would solve this specific case, but it does not scale. The renderer should stay generic, and the skill source structure should own harness-specific operational guidance.

### Additional context

Observed while using `$planbridge-open /tmp/foo.md` from Codex. The CLI logged that the annotation server was listening and that it was opening the annotation browser session, but the browser did not open automatically from the sandboxed command execution context.

blimmer:
## Summary

`contextbridge open` launches a local browser, which Codex's default sandbox blocks — `$planbridge-open` prints the listening URL but the browser never opens. The fix is content the agent reads: the Codex SKILL.md needs to tell it to set `sandbox_permissions: "require_escalated"` (with a `justification` and optionally a reusable `prefix_rule`) when invoking the command.

To deliver harness-specific content without leaking a hardcoded Codex appendix into the renderer, this PR makes SKILL.md bodies Handlebars templates with `{ harness }` as the data context and a single `eq` helper, and introduces partials under `packages/skills/sources/_partials/<harnessId>/<topic>.md`. The `codex/sandbox-escalation` partial is the first consumer; `open/SKILL.md` gains one templated block at the call site. Rendered outputs are formatted through Prettier (markdown parser) before being written, which collapses the extra blank line Handlebars leaves where a non-matching conditional sat — keeping per-harness diffs clean and letting the `harnessIntegrations/*/skills/**/SKILL.md` exclusion drop from `.prettierignore`. Closes #110.

## Review focus

- **Codex `exec_command` parameters cited in the partial are verified against the Codex source** — `sandbox_permissions: "require_escalated"`, `justification`, and `prefix_rule: ["contextbridge", "<subcommand>"]` all come from `codex-rs/core/src/tools/handlers/shell_spec.rs::create_approval_parameters`. `require_escalated` is the right mode for a browser launch (not `with_additional_permissions`, which is for additive sandbox grants).
- **Render errors throw with the skill name + harness id, with the caught error attached as `cause`** so the missing-partial Handlebars error surfaces with context: `failed to render skill 'open' for harness 'codex': The partial codex/<missing> could not be found`.
- **The partial is phrased generically** ("this skill runs commands that require resources outside Codex's default sandbox") rather than hardcoding `contextbridge open` and a specific `prefix_rule`, so a future skill wrapping a different subcommand can pull the same partial in unchanged.
- **Prettier formatting runs inside `targetsFor`** (via `prettier.resolveConfig(path)` + `prettier.format(body, { ...config, filepath: path })`) so config inherits from the destination's location in the repo. `skills:check` formats the same way, so byte-diff drift detection still works.

## Commits

- [`e66197d`](https://github.com/contextbridge/planbridge/commit/e66197d589e89b1d7cb880a982def4c686e9d095) — Handlebars templating + `_partials/` registry in `@contextbridge/skills`; `codex/sandbox-escalation` partial; `open` SKILL.md uses it; docs updated.
- [`0bc46bc`](https://github.com/contextbridge/planbridge/commit/0bc46bcdcac29eead7ea828675662b63d48e758c) — Format rendered SKILL.md through Prettier in `targetsFor`; drop the `harnessIntegrations/*/skills/**/SKILL.md` exclusion from `.prettierignore` (drift check still guards against hand-edits).
- [`1836cc9`](https://github.com/contextbridge/planbridge/commit/1836cc9fcc72689025fba3219db63a44039ebdee) — Rewrite the codex sandbox-escalation partial to drop `contextbridge open`-specific text so future skills with other commands can reuse it unchanged.
- [`56531a0`](https://github.com/contextbridge/planbridge/commit/56531a0064f52368f535781f19031c55913954f6) — PR feedback: hoist the local `toError` in `renderTargets.ts` to the shared `@contextbridge/shared/errors` implementation; replace the manual recursive `readdirSync` walk in `handlebars.ts` with `Bun.Glob('**/*.md').scanSync()`.
- [`c679f6c`](https://github.com/contextbridge/planbridge/commit/c679f6cf4581346aca84f5d3fb1fd85253ad1b21) — Inject a minimal `BaseContext` (logger + telemetry-disabled) into the `skills:generate` and `skills:check` scripts so diagnostics flow through pino instead of `console.*`.


packages/skills/scripts/check.ts
```diff
@@ -8,9 +8,9 @@ import { REPO_ROOT, type RenderTarget, SOURCES_DIR, outDirFor, targetsFor } from
 const safeReadFile = fromThrowable((path: string) => readFileSync(path, 'utf8'));
 const safeReaddir = fromThrowable((dir: string) => readdirSync(dir, { withFileTypes: true }));
 
-function main(): void {
+async function main(): Promise<void> {
```
blimmer: `async` now because we prettier-format the skills (which is an async call)

packages/skills/scripts/renderTargets.ts
```diff
@@ -13,12 +14,15 @@ export interface RenderTarget {
...
-    harness,
-    path: join(outDirFor(harness), harness.skillRendering.installName(skill.frontmatter.name), 'SKILL.md'),
-    body: render(skill, harness),
-  }));
+export async function targetsFor(skill: Skill): Promise<RenderTarget[]> {
+  return Promise.all(
+    SKILL_RENDERABLE_HARNESSES.map(async (harness) => {
+      const path = join(outDirFor(harness), harness.skillRendering.installName(skill.frontmatter.name), 'SKILL.md');
+      const config = await prettier.resolveConfig(path);
+      const body = await prettier.format(render(skill, harness), { ...config, filepath: path });
```
blimmer: The new partial was adding unneeded newlines, which prettier rightly removes.

packages/skills/sources/_partials/codex/sandbox-escalation.md
```diff
@@ -0,0 +1,3 @@
+## Running this from Codex
```
blimmer, line 1: Many skills will need this, so introducing the idea of a partial

packages/skills/AGENTS.md
```diff
@@ -19,9 +19,25 @@ A skill is `sources/<name>/SKILL.md` following the [agentskills.io specification
 
 All files are committed. `bun run skills:check` (wired into `just verify` and the lint CI job) regenerates in-memory and fails on byte-diff.
 
+## Templating
```
blimmer, line 22: I figure we'll have this with other harnesses eventually

packages/skills/src/handlebars.ts
```diff
@@ -0,0 +1,42 @@
...
+    if ((err as NodeJS.ErrnoException).code === 'ENOENT') return [];
+    throw err;
+  }
+  return entries.flatMap((entry) => {
+    const full = join(dir, entry.name);
+    if (entry.isDirectory()) return walk(full);
+    if (entry.isFile() && entry.name.endsWith('.md')) return [full];
+    return [];
+  });
+}
```
jcarver989: https://bun.com/reference/node/fs/readdir -- Bun/Node have a built in fn to recursively list files, wondering if we could use that vs rolling our own visitor/walk pattern here?

packages/skills/src/render.ts
```diff
@@ -1,13 +1,30 @@
...
+function compileBody(env: HandlebarsEnv, skill: Skill, harness: HarnessDescriptor): string {
+  try {
+    return env.compile(skill.body, { noEscape: true })({ harness });
+  } catch (err) {
+    const message = err instanceof Error ? err.message : String(err);
+    throw new Error(`failed to render skill '${skill.frontmatter.name}' for harness '${harness.id}': ${message}`, {
+      cause: err,
+    });
+  }
 }
```
jcarver989, line 36: Should this return `Result` or did you prefer the throw style ?

packages/skills/src/handlebars.ts
blimmer: Great callout! I wonder if we should have some agent skill that guides it to use more native `bun` solutions (like this, `macros` in your previous DB PR, etc.)

packages/skills/src/render.ts
blimmer: Sure we can `neverthrow` it. Do you think there's a way we can guide agents to do this? It's definitely not in their training data...

packages/skills/src/handlebars.ts
jcarver989: yea, worth trying!

packages/skills/src/render.ts
jcarver989: Yea over in `cb-bot`, ContextBot realized we do this and wrote us a rule for it here: https://github.com/contextbridge/cb-bot/blob/91dc39e5d48316db36b657365fae0482c14c3373/.contextbridge/rules/neverthrow-error-handling.md -- so I suppose the easiest thing to do would be to copy that over into our Claude rules here.

jcarver989 (approved):

jcarver989 (approved):

packages/skills/src/render.ts
blimmer: Interestingly, cb-bot did that here, too, but it's not being paid attention to. https://github.com/contextbridge/planbridge/blob/main/.contextbridge/rules/error-handling-neverthrow.md

I can't remember off the top of my head if we can edit these files? I don't think so.

packages/skills/src/handlebars.ts
blimmer: Filed as #120 — we'll handle the agent rule there.

packages/skills/src/render.ts
jcarver989: Ah interesting ... Yea currently `.contextbridge/` files are assumed to be owned by cb-bot and not-editable. 

This is a good usecase for a janitor agent that reviews the diff and fixes it w/ a fresh context window. We could either do that locally or via cb-bot on  a schedule (i.e. violation fix)

PR #116, opened 2026-05-14
feat: add global feedback submit shortcut

### LINKED ISSUE #113: feat: submit on CMD/CTRL+Enter from the overall feedback textarea
### What's the problem?

The overall feedback textarea (`GlobalCommentComposer`) in the plan review UI does not respond to keyboard shortcuts. To submit the review, users have to take their hands off the keyboard, mouse to the submit button below the textarea, and click it. That break feels heavy for a tool whose workflow is "type a thought, send it," especially since per-annotation popovers (`AnnotationPopover`) already support CMD/CTRL+Enter to save via the `useSubmitOnCmdEnter` hook — so the inconsistency is felt every time a reviewer adds an inline comment, then drops to the bottom box and reaches for the mouse.

### Proposed solution

When the overall feedback textarea (`packages/annotation/src/GlobalCommentComposer.tsx`) has focus, **CMD+Enter** (macOS) / **CTRL+Enter** (Linux/Windows) should trigger the same submit action as clicking the submit button — i.e. invoke whatever the submit button would do in its current state (the label and behavior already track state per `plan-review-design.md`'s "one contextual button" rule: it accepts when there is no feedback, requests changes when there is).

Implementation sketch:

- Reuse the existing `useSubmitOnCmdEnter` hook from `packages/annotation/src/useSubmitOnCmdEnter.ts` — same pattern used by `AnnotationPopover.tsx:43`.
- Wire `submission.submit` (the same callback used by `SubmitBar`'s `onClick`) into the textarea's `onKeyDown`.
- Respect the existing disabled state — do not fire when `submission.submitted` or `submission.submitting` is true (matches the submit button's `disabled` logic in `SubmitBar.tsx:62`).
- Add a vitest browser-mode test alongside `useSubmitOnCmdEnter.test.ts` / the existing App tests covering: (a) CMD+Enter submits when textarea is focused, (b) CTRL+Enter submits when textarea is focused, (c) no submit while `submitting`/`submitted`, (d) plain Enter still inserts a newline.

The shortcut should only fire while the textarea is focused (per the user's request) — no global document-level listener.

- **Global hotkey** that submits regardless of focus. Rejected: surprises users who are typing inside an inline annotation popover, where CMD+Enter already means "save this annotation."
- **Plain Enter to submit** (Slack/iMessage style). Rejected: the textarea explicitly supports multi-line feedback and the placeholder ("Add overall feedback…") invites prose — stealing Enter for submit breaks paragraph entry.

### Additional context

Today's CMD+Enter wiring for reference:

- Hook: `packages/annotation/src/useSubmitOnCmdEnter.ts`
- Existing consumer: `packages/annotation/src/AnnotationPopover.tsx:43,108`
- Target component: `packages/annotation/src/GlobalCommentComposer.tsx`
- Submit pipeline (the callback to invoke): `SubmissionState.submit` in `packages/annotation/src/SubmitBar.tsx:12`

blimmer:
## Summary

Adds CMD/CTRL+Enter submission support to the overall feedback textarea, reusing the existing textarea-scoped shortcut hook and the same submission path as the contextual submit button. The shortcut respects submitting/submitted guards, and browser coverage now exercises the new keyboard path plus newline behavior.

Closes #113.

## Review focus

Please focus on the shortcut wiring between `GlobalCommentComposer` and `SubmitBar`'s submission state, especially that the handler stays textarea-scoped and does not conflict with inline annotation popovers. Also worth checking: the new shared `createDeferred` test helper and the added testing guidance for future agents.

## Commits

- [`51230e1`](https://github.com/contextbridge/planbridge/commit/51230e151a273e681960aabf9fbe8e4857d114a2) — add global feedback keyboard submit support and tests.


.contextbridge/rules/testing-patterns.md
```diff
@@ -24,7 +24,11 @@ globs: ["packages/**/*.test.ts", "packages/**/*.test.tsx", "packages/**/*TestIds
   const env = environment.build({ HOME: tmp });
   ```
 
-- **Export component test IDs as a colocated object from the component file.** Each React component that exposes test selectors declares its own \`xxxTestIds\` object next to the component definition and exports it. Tests import this object directly, never using string literals for test IDs. The naming pattern is \`\<componentName>TestIds\`.
+- **Prefer `toMatchObject` for structured payload assertions.** When a test verifies several fields on the same object or nested payload, use one `expect(value).toMatchObject({ ... })` instead of a run of field-by-field assertions. Keep separate assertions for orthogonal behavior, clearer failure messages, or values that need a specialized matcher.
+
+- **Use the shared deferred-promise helper; do not hand-roll it.** Tests that need manual promise resolution should import `createDeferred` from `@contextbridge/shared/testHelpers`. Do not recreate local `Deferred` types or `new Promise` wrappers in individual test files.
```
blimmer, line 29: Trying to guide the agent to quit doing this

packages/shared/src/testHelpers.test.ts
```diff
@@ -0,0 +1,21 @@
+import { describe, expect, it } from 'bun:test';
+import { createDeferred } from './testHelpers.ts';
+
+describe('createDeferred', () => {
```
blimmer, line 4: Hoisting this

jcarver989 (approved):

PR #119, opened 2026-05-14
chore: move public docs website from private tools submodule

### LINKED ISSUE #99: chore: move public docs website from private tools submodule
Contributors should be able to easily make changes to the astro website, but can't right now because of the submodule. And, in general, the submodule is janky even for us.

Notes:
- It's OK to eject from projen, since our projen module is currently closed source.
- We should s3 sync via a new OIDC role (defined in private infrastructure)
- The website should only deploy on stable release (so unreleased features are not documented before they're available)

blimmer:
## Summary

Moves the Astro + Starlight marketing/docs site out of the private `planbridge-private` tools submodule and into this repo as a regular `@contextbridge/website` workspace under `packages/website`, so public contributors can change the site without any submodule setup. Strips the projen-generated infrastructure that only existed to manage that submodule, and adds a `deploy-website` job to the release workflow that publishes the built site to S3 + CloudFront on stable releases only (alpha tags are skipped so unreleased docs do not publish before the corresponding CLI release).

## Review focus

- The CI deploy job assumes three new bits of GitHub config: `vars.WEBSITE_BUCKET`, `vars.WEBSITE_DISTRIBUTION_ID`, and `secrets.WEBSITE_DEPLOY_ROLE_ARN`. The OIDC role is defined in private infrastructure (per #99) and must exist before the next stable release tag, otherwise the new job will fail. The job runs only when `!contains(github.ref_name, '-')` — worth double-checking that matches how release tags are shaped.
- Cache-control split: `_astro/*` gets `public, max-age=31536000, immutable` (fingerprinted assets), everything else gets `public, max-age=300, must-revalidate`. Invalidation hits `/*` regardless, so the immutable cache only matters for in-flight clients between deploys.
- The generated `packages/website/src/content/docs/cli/` reference is gitignored. It is regenerated by Astro lifecycle hooks (`predev`/`prebuild`/`pretypecheck`) from `packages/cli/scripts/genCliRef.ts`. The path move is also reflected in `packages/annotation/scripts/recordDemo.ts` (homepage demo asset output).
- Dependabot's `actions/create-github-app-token` ignore was projen-specific and is dropped — dependabot will resume managing that pin directly.

## Commits

- [`44d35c1`](https://github.com/contextbridge/planbridge/commit/44d35c1) — chore: drop projen + private tools submodule plumbing
- [`af314cb`](https://github.com/contextbridge/planbridge/commit/af314cb) — feat(website): add public packages/website workspace (#99)
- [`07885bc`](https://github.com/contextbridge/planbridge/commit/07885bc) — ci(release): deploy website to S3/CloudFront on stable release

Closes #99.


.github/workflows/auto-approve.yml
blimmer, line 1: This was old / from projen

.github/workflows/deploy-website.yml
```diff
@@ -0,0 +1,72 @@
+name: Deploy Website
```
blimmer, line 1: Reusable workflow used from `main` and `release` workflows

.tool-versions
```diff
@@ -1,2 +1,3 @@
 bun 1.3.13
 just 1.49.0
+nodejs 24.15.0
```
blimmer, line 3: `astro check` uses a shebang for `node`, so we need this, unfortunately. They explicitly call out bun as out of scope for `check`.

PR #127, opened 2026-05-14
ci: pin website CLI reference to stable tags

### LINKED ISSUE #124: feat: pin website CLI reference docs to latest stable release tag
### What's the problem?

The website's auto-generated CLI reference (`packages/cli/scripts/genCliRef.ts`) introspects the Commander program from whatever `packages/cli/src/` is currently checked out. Today on `main` that means the published marketing site can describe commands or flags that don't yet exist in any released CLI binary — the docs drift ahead of the binary every time something lands on `main` between releases.

The recently-merged decoupling of the website deploy from the release workflow (reusable `deploy-website.yml`, path-filtered on `main`, full deploy on stable release) makes this more visible: marketing copy edits now ship within minutes of merging to `main`, and so does any unreleased CLI surface that happened to land alongside them.

### Proposed solution

Add an optional `cli_ref` input to `.github/workflows/deploy-website.yml`. Callers pass:

- `main.yml`: the latest stable tag resolved from `git tag --list 'v*' --sort=-version:refname` filtered to non-prerelease (no `-` in the name).
- `release.yml`: `github.ref_name` (the tag being released).

Inside the reusable workflow, before `bun run --cwd packages/website build`:

1. `git fetch --tags --no-tags origin "$CLI_REF"` (or equivalent — make sure the tag is in the local repo).
2. `git checkout "$CLI_REF" -- packages/cli`.
3. Let the existing `prebuild` hook run `gen:docs` against the now-tagged `packages/cli/`.

Pass `CLI_REF` through run blocks as `env:` (not `${{ }}` interpolation) so zizmor's template-injection audit stays clean — tag names are mostly trusted but the input is technically caller-controlled.

### Additional context

- Depends on the reusable-workflow PR landing first (today's deploy builds from `main` and ignores tag pinning).
- No backward-compatibility shim is needed: the next stable release cut after that PR merges becomes the floor for this work. From that release forward, the tagged `genCliRef.ts` already writes to `packages/website/src/content/docs/cli/`, so a straight `git checkout` is sufficient. Pre-existing tags (v0.5.0 and earlier) predate the workspace restructure and aren't in scope.
- Optional adjunct worth considering alongside: teach `genCliRef.ts` to honor a `CLI_DOCS_OUT_DIR` env var so future structural moves don't require re-cutting a release before pinning works again. Not load-bearing for this issue, but cheap insurance.

blimmer:
## Summary

Pins the website CLI reference generation to released CLI source instead of whatever `packages/cli` happens to be on `main`. Main-branch website deploys now resolve the latest stable `v*` tag, while stable release deploys pass the release tag itself, so the generated docs stay aligned with published binaries. Closes #124.

## Review focus

Please sanity-check the workflow ref handling: caller-controlled refs are passed through `env`, validated as stable `v*` tags, fetched by exact tag refspec, and resolved with `^{commit}` before checking out `packages/cli`.

## Commits

- [`70f26b0`](https://github.com/contextbridge/planbridge/commit/70f26b07a743b218b300d3ac2167f73312ff3b47) — pin website CLI reference generation to stable tags

### DIFF
--- a/.github/workflows/deploy-website.yml
+++ b/.github/workflows/deploy-website.yml
@@ -2,6 +2,12 @@ name: Deploy Website
 
 on:
   workflow_call:
+    inputs:
+      cli_ref:
+        description: Stable CLI tag to use when generating website CLI reference docs. Empty means use the checked-out CLI source.
+        required: false
+        type: string
+        default: ''
     secrets:
       website_deploy_role_arn:
         description: ARN of the OIDC role to assume for the S3 sync and CloudFront invalidation.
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-website.yml
```diff
@@ -38,6 +44,40 @@ jobs:
...
+            *-*)
+              echo "::error::CLI_REF must be a stable tag without '-', got '${CLI_REF}'"
+              exit 1
+              ;;
+          esac
+
+          git fetch --force --no-tags --depth=1 origin "refs/tags/${CLI_REF}:refs/tags/${CLI_REF}"
+          cli_ref_commit="$(git rev-parse --verify "refs/tags/${CLI_REF}^{commit}")"
+          git checkout "${cli_ref_commit}" -- packages/cli
+          echo "Pinned packages/cli to ${CLI_REF} (${cli_ref_commit})"
```
blimmer, line 79: @jcarver989 - our idea to only copy over the cli docs in the two workflows doesn't quite work. The issue is the sidebar, which is included on all the pages. So a new command would show up in the sidebar but then 404.

jcarver989: ah, yea makes sense.

jcarver989 (approved):

PR #128, opened 2026-05-14
fix: quiet post-update harness refresh

blimmer:
## Summary

`contextbridge update` now refreshes installed harness integrations with `contextbridge install <harness> --quiet` after a successful self-update, then prints one concise refresh line per harness. This keeps the useful post-update confirmation while suppressing first-install onboarding prose that is noisy and misleading during an update refresh.

The PR also extracts shared IO behavior into `BaseIo` so `IoImpl` and `FakeIo` do not duplicate quiet-aware writes, stdin reads, and TTY handling.

## Before

```text
$ contextbridge update
A new version is available: v0.6.0-alpha.0 (you're on v0.5.0-alpha.3).
What's new: https://github.com/contextbridge/planbridge/releases/tag/v0.6.0-alpha.0
...
✓ PlanBridge plugin installed for Claude Code (scope: user).
Restart Claude Code for the plugin to load.
✓ PlanBridge hook installed for Codex CLI (scope: user).
Action required: PlanBridge will not run in Codex until this hook is trusted.
Restart Codex CLI, open /hooks, verify the Stop hook runs `contextbridge hook codex`, and press t.
Walkthrough: https://plan.contextbridge.ai/usage/codex/#trust-the-codex-hook
✓ PlanBridge skill installed at ~/.agents/skills/planbridge-open/SKILL.md.
Invoke in Codex as $planbridge-open.
✓ update complete.
```

## After

```text
$ contextbridge update
A new version is available: v0.6.0-alpha.0 (you're on v0.5.0-alpha.3).
What's new: https://github.com/contextbridge/planbridge/releases/tag/v0.6.0-alpha.0
...
✓ refreshed Claude Code (scope: user).
✓ refreshed Codex CLI (scope: user).
✓ update complete.
```

## Review focus

Please sanity-check the hidden `--quiet` installer option and the `IoImpl.from(ctx.io, { quiet })` context wrapping in `ScopedHarnessInstaller`. The goal is to suppress only installer prose while preserving raw inherited stdio for command runner passthrough and parent command summaries.

## Commits

- `f3e3339` — quiet post-update harness refresh


packages/cli/src/testHelpers/FakeIo.ts
```diff
@@ -1,32 +1,13 @@
 import { PassThrough } from 'node:stream';
-import { type Io } from '#src/IoImpl.ts';
-import { readStreamToString } from '#src/streams.ts';
+import { BaseIo } from '#src/BaseIo.ts';
 import { MemoryStream } from './MemoryStream.ts';
 
-export class FakeIo implements Io {
```
blimmer, line 6: I was duplicating a lot of logic with the new `quiet` flag, so I used an abstract class instead.

jcarver989 (approved):
Thanks, makes sense to use a base class to avoid duplication here.

PR #141, opened 2026-05-15
docs(website): add Superpowers recipe

blimmer:
## Summary

Adds a new "Recipes" sidebar group to the docs site whose first entry documents how to pair PlanBridge with the obra/superpowers Claude Code / Codex plugin. The recipe covers the two pause points where inline human review fits cleanly into superpowers' workflow: mid-brainstorm (annotating a design section while it is still in the terminal) and post-spec (annotating the spec file once the brainstorming skill writes it to disk).

Also trims the inline parenthetical in `usage/open.mdx` that hand-explained superpowers' plan-mode bypass, replacing it with a one-line pointer to the new recipe so the same content does not live in two places.

## Review focus

The IA choice. "Recipes" is a new top-level sidebar group rather than a subsection of "Usage". The framing is that this group is for tools PlanBridge pairs with but does not directly integrate with, leaving room for future entries (exe.dev, etc.) without polluting the Usage group with third-party-specific pages. Open to renaming the group if a better label surfaces in review.

## Commits

- [`0d26289`](https://github.com/contextbridge/planbridge/commit/0d26289) — docs(website): add Superpowers recipe page


packages/website/src/content/docs/recipes/superpowers.mdx
```diff
@@ -0,0 +1,51 @@
+---
```
blimmer, line 1: cc @floodfx - I think `recipes` could be a good landing place for exe.dev details, as well.

blimmer:
I have a few superpowers users I want to reach out to before my EOD, so I'm going to merge as-is. Post-merge feedback is welcomed!

PR #142, opened 2026-05-15
chore(agents): Fix Dockerfile which required node for agents to run a…

jcarver989:
This PR fixes some issues with our agent's Dockerfile, changes are in response to an AWS error. Previously we were installing an old version of node (18), but we need a newer version to avoid runtime errors that depend on `AsyncDisposableStack`

### DIFF
--- a/.contextbridge/Dockerfile
+++ b/.contextbridge/Dockerfile
@@ -10,11 +10,10 @@ FROM debian:bookworm-slim
 ARG AGENT_UID=1000
 ARG AGENT_GID=1000
 ARG ASDF_VERSION=0.18.0
-ARG NODE_VERSION=24.15.0
 
 # Chromium runtime libs drive @contextbridge/annotation's vitest --browser tests.
 # libdbus-1-3 is required by aether's native ACP subprocess.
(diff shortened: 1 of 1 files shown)


.contextbridge/Dockerfile
```diff
@@ -1,37 +1,34 @@
...
 #   protocol script at /workspace/agent/customer-entrypoint.sh, so this image
 #   intentionally has no ENTRYPOINT/CMD of its own
 
-FROM oven/bun:1.3.13-debian
-ARG JUST_VERSION=1.49.0
+FROM debian:bookworm-slim
 ARG AGENT_UID=1000
 ARG AGENT_GID=1000
+ARG ASDF_VERSION=0.18.0
+ARG NODE_VERSION=24.15.0
```
blimmer: Node is managed by ASDF - do we need this?

blimmer (approved):

.contextbridge/Dockerfile
jcarver989: Oversight on my part, I had thought only bun was in there but misremembered.

blimmer: Just added it in the last ~36 hours!

PR #165, opened 2026-05-18
fix: render images referenced from the local filesystem

### LINKED ISSUE #148: Plan images render as broken localhost file URLs
Images in plans do not render properly. The generated image URLs are local file paths that the local CLI does not resolve correctly.

Example path:<br>`/Users/<user>/.codex/generated_images/019e2d6b-3f30-71d2-b5d1-9682deb6029f/ig_08af568a4f50a63a016a078b0698ec8199aa3037dc5f7b7eb6.png`

This becomes a localhost URL instead of a usable image URL.

blimmer:
## Summary

Plan markdown containing local image references rendered as broken images because the browser tried to resolve filesystem paths against the localhost annotation server. This PR extracts local, allowlisted image files at annotation startup, embeds their bytes as JSON-safe base64 assets on `AnnotationPayload`, and serves those embedded assets through `/assets/:id` for the current browser UI. The payload is now portable enough for future shareable plan links while the markdown content remains byte-identical end-to-end. Closes #148.

## Review focus

Please scrutinize the asset contract and security boundary: image bytes now live in the payload as `dataBase64` plus `mimeType`, and the server only serves assets by opaque ID from the in-memory payload map. There are no request-time filesystem reads and no network/data-URL fetching during extraction.

The extraction limits are also deliberate: local images are capped at 10 MiB each and 25 MiB total, SVG remains excluded, and unsupported/empty/oversized images are skipped with logs instead of failing the review session. That keeps accidental huge/private payloads bounded, but it also means a skipped local image can still render as the original unresolved markdown path.

## Commits

Stack is ordered by layer for commit-by-commit review: shared payload contract → server serving → browser rendering → CLI extraction/inlining.

- [`2038ab1`](https://github.com/contextbridge/planbridge/commit/2038ab1de1c1b1e62f5e1b08dfd3e5cccbc85899) — feat(shared): add portable annotation assets
- [`d80053a`](https://github.com/contextbridge/planbridge/commit/d80053a0efe1c4436781b4b041a98e984db747e3) — feat(server): serve embedded annotation assets
- [`6069173`](https://github.com/contextbridge/planbridge/commit/6069173d9598bd3927884de3676844fba019b06a) — feat(annotation): render payload image assets
- [`cf16ccf`](https://github.com/contextbridge/planbridge/commit/cf16ccfefb2564958e10b20fedfc0b3159f87791) — fix(cli): inline local markdown images as assets


blimmer:
I've gotta run to my lunch meeting w/ Ron so didn't have time to rebase the commits. Feel free to look now, or wait until I'm back to have an agent rebase

packages/annotation/src/AnnotatedMarkdown.tsx
```diff
@@ -14,9 +14,23 @@ export interface AnnotatedMarkdownProps {
...
-export function AnnotatedMarkdown({ content, containerRef, onMouseUp }: AnnotatedMarkdownProps) {
+export function AnnotatedMarkdown({ content, containerRef, onMouseUp, assets }: AnnotatedMarkdownProps) {
+  const assetsByPath = new Map<string, Asset>();
+  for (const asset of assets ?? []) {
+    assetsByPath.set(asset.originalPath, asset);
+  }
+
+  const components = {
+    ...markdownComponents,
+    img: ({ src, alt, ...rest }: ComponentPropsWithoutRef<'img'>) => {
```
blimmer, line 28: Snags images defined in our assets metadata and serves them from the purpose-built CLI endpoint.

packages/server/src/annotation.ts
```diff
@@ -104,6 +110,19 @@ export function createAnnotationServerApp(ctx: ServerContext, opts: StartServerO
         setTimeout(() => resolveResult(parsed.data), 0);
         return new Response(null, { status: 204, headers: { connection: 'close' } });
       }
+      if (req.method === 'GET' && url.pathname.startsWith('/assets/')) {
```
blimmer: This method is getting gnarly AF - see https://github.com/contextbridge/planbridge/pull/170 for a refactor

packages/cli/src/annotation/extractAssets.ts
```diff
@@ -0,0 +1,167 @@
...
+import type { CliContext } from '#src/context.ts';
+
+export interface ExtractAssetsInput {
+  content: string;
+  sourcePath?: string;
+}
+
+const MIB = 1024 * 1024;
+export const MAX_ASSET_BYTES = 10 * MIB;
+export const MAX_TOTAL_ASSET_BYTES = 25 * MIB;
```
blimmer, line 21: I don't expect this large of images, but I figured I'd have some safety mechanism since we base64 encode the data.

packages/cli/src/annotation/extractAssets.ts
```diff
@@ -0,0 +1,250 @@
...
+import { fileURLToPath } from 'node:url';
+import { ASSET_FILE_EXTENSIONS, type Asset, AssetMimeTypeSchema } from '@contextbridge/shared/annotationSchema';
+import type { Image } from 'mdast';
+import { Result, ResultAsync } from 'neverthrow';
+import remarkParse from 'remark-parse';
+import { unified } from 'unified';
+import { visit } from 'unist-util-visit';
+import type { CliContext } from '#src/context.ts';
+
+export interface ExtractAssetsInput {
```
jcarver989, line 14: Ah yea I see what you were talking about know in our earlier discussion. Some observations/thoughts here: 

### 1. 
I'm not sure if you'd find this nicer, but the current code is doing a lot of check x and log y + return `null` for things that are essentially errors (or perhaps more accurately rejections), so you might consider creating a type like: 

```typescript
export type AssetExtractionIssue = IssueA | IssueB | IssueC...
````
That would allow several of these functions to return a `Result` and you could log only at the top level. 

### 2. 

You might consider a `AssetFilter` type that looks something like: `type AssetFilter = (candidate) => Result<void, AssetExtractionIssue>

That'd allow you to compose different filter rules together more easily,

jcarver989 (approved):
Left some thoughts on cleaning up the big function you mentioned, but nothing blocking.

packages/cli/src/annotation/extractAssets.ts
blimmer: I like it - thanks

PR #169, opened 2026-05-18
docs(website): add automatic wiring to Superpowers recipe

### LINKED ISSUE #146: investigation: superpower workflow "remember"
At one point I said something like "always open brainstorming sections and specs in planbridge" during a superpowers session. It seems to have remembered this and now it "just works" in all my superpowers sessions.

We should dig into that more. If there's a good prompt we could give people, that might be easier than what we document in the [superpowers recipe](https://plan.contextbridge.ai/recipes/superpowers/) today!

relates to #143

blimmer:
## Summary

Splits the Superpowers recipe into Manual and Automatic sections. The Automatic path is a single `echo` per harness that appends a prompt referencing `/planbridge-open` (`$planbridge-open` for Codex) so both brainstorming design sections and spec review route through PlanBridge without per-pause user input. Closes #146.

## Review focus

The Automatic snippet is an unbounded append to `~/.claude/CLAUDE.md` / `~/.codex/AGENTS.md` — no sentinel markers, no idempotency. Worth deciding whether the docs should call that out, or whether the right follow-up is wiring this through `contextbridge install` so it's reversible.

## Commits

- [`3d0559f`](https://github.com/contextbridge/planbridge/commit/3d0559ff2e7bdebaa67801d525980f1f90bdc8d2) — docs(website): add automatic wiring to Superpowers recipe

### DIFF
--- a/packages/website/src/content/docs/recipes/superpowers.mdx
+++ b/packages/website/src/content/docs/recipes/superpowers.mdx
@@ -8,14 +8,17 @@ import { Code } from 'astro-expressive-code/components';
 
 [Superpowers](https://github.com/obra/superpowers) is a Claude Code and Codex plugin that gives the agent a library of process skills (brainstorming, writing plans, executing plans, debugging).
 
-It explicitly tells the agent to not use Plan mode, which breaks our default behavior. However, there are two pause points in that workflow that are good places to utilize PlanBridge for review: brainstorming and spec review.
+There are two pause points in the Superpowers workflow where PlanBridge fits well: design sections during brainstorming, and spec review. You can trigger PlanBridge manually or automatically.
 
-## During brainstorming
+## Manual
 
-The brainstorming skill walks the agent through clarifying questions and presents the design back to you in sections, in the terminal.
-Often, a section might have hundreds of lines of output, making it difficult to provide targeted feedback.
+Trigger PlanBridge yourself with the `planbridge-open` slash command. This allows you to respond to some superpowers sections (like short brainstorming sections) in the terminal UI, while opening larger sections in PlanBridge to provide more precise feedback.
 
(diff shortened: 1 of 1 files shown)


packages/website/src/content/docs/recipes/superpowers.mdx
```diff
@@ -8,14 +8,17 @@ import { Code } from 'astro-expressive-code/components';
 
 [Superpowers](https://github.com/obra/superpowers) is a Claude Code and Codex plugin that gives the agent a library of process skills (brainstorming, writing plans, executing plans, debugging).
 
-It explicitly tells the agent to not use Plan mode, which breaks our default behavior. However, there are two pause points in that workflow that are good places to utilize PlanBridge for review: brainstorming and spec review.
+There are two pause points in the Superpowers workflow where PlanBridge fits well: design sections during brainstorming, and spec review. You can trigger PlanBridge manually or automatically.
 
-## During brainstorming
+## Manual
```
blimmer, line 13: I wanted to emphasize that you can trigger it manually (what we had documented before) or automatically (new)

packages/website/src/content/docs/recipes/superpowers.mdx
```diff
@@ -44,6 +46,43 @@ another great time to use PlanBridge to review and annotate the spec. A better s
 
 Same loop: annotate inline, submit, and the agent revises the spec or moves on to writing the plan.
 
+## Automatic
```
blimmer, line 49: This felt like something simple we could do now. Alternatively, we could ship a skill that tries to do this automatically and that we manage through our existing workflow. Thoughts?

jcarver989: Think this seems ok for now.

jcarver989 (approved):

PR #170, opened 2026-05-18
refactor(server): split annotation routes via Bun routes API

blimmer:
## Summary

The annotation server's `fetch` handler was a long if/else chain branching on `req.method` and `url.pathname` — felt messy and didn't scale to the upcoming `review` surface. This swaps it for Bun's declarative `Bun.serve({ routes })` API and splits each handler into its own file under `packages/server/src/routes/`, so `annotation.ts` becomes a thin composition layer (routes table + lifecycle) and new routes drop in beside the existing ones.

Tests follow the split: per-route HTTP integration tests live colocated as `routes/<name>.test.ts` and share a `withServer(ctx, opts?, fn)` helper that defaults html/payload/config via new Fishery factories. `annotation.test.ts` keeps smoke + lifecycle coverage (404 fallback, `resolveListenPort`, and the existing submit→close race regression).

## Review focus

- **`/submit` invariants** — the `setTimeout(() => resolveResult(...), 0)` deferral and `connection: close` header are non-negotiable (they prevent Bun from tearing down the socket before the 204 is flushed). They moved into `routes/submit.ts` verbatim and the race regression test in `annotation.test.ts` still passes, but worth a second pair of eyes.
- **Test surface tradeoff** — `app.fetch(req)` in-process is gone; everything is real HTTP against a started server now. Slightly slower per test, but a stronger integration boundary and removes the `createAnnotationServerApp` factory whose only consumer was the tests themselves.
- **`withServer` overload** — opted for two signatures (`(ctx, fn)` vs `(ctx, opts, fn)`) over a required `{}` middle arg. The overload impl is 4 lines of `typeof optsOrFn === 'function'` — let me know if you'd prefer the simpler always-3-arg shape.

## Commits

- 6754e35 — refactor(server): split annotation routes via Bun routes API

### DIFF
--- a/packages/server/src/annotation.test.ts
+++ b/packages/server/src/annotation.test.ts
@@ -1,200 +1,10 @@
 import { fakeBaseContext } from '@contextbridge/context/testHelpers';
-import type { AnnotationPayload } from '@contextbridge/shared/annotationSchema';
-import type { FrontendConfig } from '@contextbridge/shared/frontendConfigSchema';
-import { annotationThread, globalThread } from '@contextbridge/shared/testFactories';
-import type { UpdateNotice } from '@contextbridge/shared/updateNoticeSchema';
+import { annotationPayload, frontendConfig } from '@contextbridge/shared/testFactories';
 import { describe, expect, it } from 'bun:test';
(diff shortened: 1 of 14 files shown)


jcarver989 (approved):
Looks like a nice cleanup here thanks

PR #188, opened 2026-05-20
fix: prevent data loss with onbeforeunload warning

blimmer:
## Summary

Adds a custom close-review dialog when a user attempts to leave an unsubmitted annotation review. The dialog now recommends approving an empty review or submitting existing feedback, while keeping the browser beforeunload guard in place.

### No comments

Screenshot 2026-05-20 at 09 55 44

### With comments

Screenshot 2026-05-20 at 09 56 07

## Review focus

Please look closely at the `FrontendBrowser.addBeforeUnloadGuard` callback shape and whether triggering React dialog state from the beforeunload path is the right API boundary.

## Commits

- [`75f4d32`](https://github.com/contextbridge/planbridge/commit/75f4d32) — prompt before closing annotation review

### DIFF
--- a/packages/annotation/src/App.test.tsx
+++ b/packages/annotation/src/App.test.tsx
@@ -3,7 +3,7 @@ import { DOCS_URL, FEEDBACK_URL, GITHUB_REPO_URL, SLACK_COMMUNITY_URL } from '@c
 import { createDeferred } from '@contextbridge/shared/testHelpers';
 import type { UpdateNotice } from '@contextbridge/shared/updateNoticeSchema';
 import { headerTestIds } from '@contextbridge/ui/components/Header';
-import { act, cleanup, fireEvent, screen, waitFor } from '@testing-library/react';
+import { act, cleanup, fireEvent, screen, waitFor, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 import { annotatedMarkdownTestIds } from './AnnotatedMarkdown.tsx';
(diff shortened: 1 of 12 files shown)


packages/annotation/src/App.test.tsx
```diff
@@ -157,6 +157,63 @@ describe('App', () => {
...
+    });
+
+    const dialog = await screen.findByTestId(appTestIds.closeReviewDialog);
+    expect(dialog).toHaveTextContent('Approve plan before closing?');
+    expect(dialog).toHaveTextContent(
+      'No comments have been added. Select Approve Plan to tell the agent to proceed with the plan as written.',
+    );
+    expect(within(dialog).getByRole('button', { name: 'Keep Reviewing' })).toBeInTheDocument();
+    expect(within(dialog).getByRole('button', { name: 'Approve Plan' })).toBeInTheDocument();
+    expect(within(dialog).queryByRole('button', { name: 'Close Anyway' })).not.toBeInTheDocument();
```
blimmer: Review agent: use test ids throughout vs. getByRole. Review the agent instructions for details.

packages/annotation/src/App.test.tsx
```diff
@@ -157,6 +157,63 @@ describe('App', () => {
     });
   });
 
+  it('opens the approval close dialog after staying on an attempted close with no comments', async () => {
```
blimmer, line 161: This helps prevent the case described here: https://contextbridge.slack.com/archives/C0ATDN9SSR3/p1779224919519899

Where a user closed the tab without clicking approve. I learned that browsers do require some interaction with the page before `onbeforeunload` will work (to avoid spammy stuff). But I assume most people will have at least scrolled the document.

packages/annotation/src/useAnnotationState.test.ts
```diff
@@ -85,6 +86,113 @@ describe('useAnnotationState', () => {
     });
   });
 
+  describe('beforeunload guard', () => {
```
blimmer, line 89: I tested these cases manually and they all work.

packages/annotation/src/App.test.tsx
```diff
@@ -157,6 +157,63 @@ describe('App', () => {
...
+
+    await act(async () => {
+      browser.triggerBeforeUnload();
+      await Promise.resolve();
+    });
+
+    const dialog = await screen.findByTestId(appTestIds.closeReviewDialog);
+    expect(dialog).toHaveTextContent('Approve plan before closing?');
+    expect(dialog).toHaveTextContent(
+      'No comments have been added. Select Approve Plan to tell the agent to proceed with the plan as written.',
```
jcarver989: These sorts of tests are super brittle because they test copy that can change at any time. I'd either just assert that the test id you want is there or alternatively pull the copy into a constant / exported var and import it in both the component and tests.

jcarver989 (approved):
Overall looks good, would like to make these tests less brittle before merging though.

packages/annotation/src/App.test.tsx
blimmer: Historically I would have agreed strongly with you. But idk that I feel that way with AI-driven development. If the message changes, `just verify` will fail and the agent will just update the test.

My main goal here is that we're testing the two different cases, which it feels like we're doing here decently well right now. But if you prefer the constant, that's an easy fix :+1:

jcarver989: You still end up burning a lot of wall clock time and tokens while the agent fumbles around fixing the tests. And that scales as an O(N) problem the more brittle tests the agent adds. 

I think it's good to do it now so it re-enforces the pattern with agents to not do that, which avoids them needing to update 100 broken tests b/c we changed copy when our test suite gets huge later down the road.

blimmer: I'll extract to constants

Issue #182, opened 2026-05-19
feat: ship a /planbridge-last skill to annotate the agent's most recent message


blimmer:
### What's the problem?

One of the highest-leverage uses of PlanBridge in practice is: the agent just produced a wall of text (an architecture explanation, a proposed migration, a draft commit message, a code-review summary, a brainstorming section), and you want to mark it up inline rather than typing prose corrections back into the terminal.

Today this is achievable, but only as an undocumented trick: `/planbridge-open the last thing you said` (or similar plain-language phrasing) and hope the agent resolves it correctly. That works often enough to be useful, but it depends on:

1. The user knowing this prompt pattern exists at all (no discoverability — it's not in the autocomplete picker).
2. The agent interpreting the temporal reference consistently (varies by harness, varies by model version).
3. The agent copying its prior turn **verbatim** rather than paraphrasing or summarizing (the default failure mode without explicit instruction).
4. The agent not emitting a preamble like "Sure, let me open that for you" *before* invoking the command — because that preamble becomes the new "last thing."

Two of these are fixable in a docs recipe. The other two (discoverability + reliable agent behavior) are not.

### Proposed solution

Ship a dedicated `/planbridge-last` skill in the `planbridge@contextbridge` plugin, parallel to `/planbridge-open`. Its job: grab the agent's most recent assistant message and open it in PlanBridge for annotation.

**Mechanism.** The skill instructs the agent to pipe its prior turn verbatim to `contextbridge open` via stdin. No new CLI subcommand needed — `contextbridge open` already accepts piped content. The skill is a thin SKILL.md that tells the agent exactly what to do.

**Source location.** `packages/skills/sources/planbridge-last/SKILL.md`. Renders to `harnessIntegrations/claude/skills/planbridge-last/SKILL.md` and `harnessIntegrations/codex/skills/planbridge-last/SKILL.md` via `bun run skills:generate`. Register in `packages/skills/src/codex.ts` so the CLI embeds it for Codex install. Run `bun run skills:check && bun run --cwd packages/cli test CodexInstaller` per the standard skill-add flow.

**Proposed SKILL.md.** Mirrors the structure of `planbridge-open` (when-to-use / how-to-invoke / what-happens / what-to-do-with-output / limitations) so the two skills feel like siblings:

````markdown
---
name: planbridge-last
description: Open your most recent assistant message in the PlanBridge browser UI for human annotation. Use when the user wants to annotate or give feedback on the wall of text you just wrote.
---

[...]

PR #189, opened 2026-05-20
feat: add planbridge-last skill to open the last agent message

blimmer:
## Summary

Adds the `planbridge-last` skill so Claude Code and Codex users can open the agent previous assistant message in PlanBridge for annotation. It also registers the bundled Codex skill, broadens installer tests to cover all bundled skills, and updates usage docs to point users at the new manual flow. References #182.

## Review focus

Please look closely at the skill wording around verbatim reproduction and preamble suppression, since that is the behavior this feature depends on across harnesses.

## Commits

- [`457550e`](https://github.com/contextbridge/planbridge/commit/457550e8e576530b7928babff81c96598210bda5) — add planbridge-last skill

### DIFF
--- a/packages/skills/sources/planbridge-last/SKILL.md
+++ b/packages/skills/sources/planbridge-last/SKILL.md
@@ -20,15 +20,23 @@ Use this skill when the user wants to mark up something you just said. Typical t
 
 **Do not send any commentary or status message before running the command.** The command targets your most recent rendered message, so a preamble like "Sure, opening that now" can mistakenly become the thing being annotated. Run the command first; speak after.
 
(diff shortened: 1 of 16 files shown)


packages/skills/sources/planbridge-last/SKILL.md
```diff
@@ -0,0 +1,58 @@
+---
```
blimmer, line 1: Feedback welcome on the phrasing here!

jcarver989 (approved):
Looks good to me. Refraining on commenting about prompt suggestions as we do not have evals in place at this point and so anything I say would be purely conjecture / my own personal biases.

PR #191, opened 2026-05-20
refactor: extract user-visible copy into colocated constants

blimmer:
## Summary

Aligns three existing UI surfaces with the test-copy-constants rule introduced in #188 (`.claude/rules/test-copy-constants.md`): `UpdateNoticeCard` text, the close-review dialog's "Keep Reviewing" cancel label, and the empty-state notice now flow through exported `xxxCopy` objects that the corresponding tests import. Also removes two stale `not.toHaveTextContent('Close Anyway')` assertions whose target string never existed in source.

## Review focus

The cancel label is shared by both close-review variants, so I added it as a top-level `closeReviewDialogCopy.cancelLabel` and surfaced it on the hook's `closeReview` return object alongside the variant-driven `title` / `description` / `primaryActionLabel`. The alternative was importing `closeReviewDialogCopy` directly in `App.tsx` and bypassing the hook for the static piece — I went with the hook to keep all dialog copy flowing through one seam, but it's worth a sanity check.

## Commits

- [`a3ce14e`](https://github.com/contextbridge/planbridge/commit/a3ce14eadc16a7d94bd760435faca4dc6788a72a) — refactor: extract user-visible copy into colocated constants

### DIFF
--- a/packages/annotation/src/App.test.tsx
+++ b/packages/annotation/src/App.test.tsx
@@ -8,7 +8,7 @@ import userEvent from '@testing-library/user-event';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 import { annotatedMarkdownTestIds } from './AnnotatedMarkdown.tsx';
 import { annotationPopoverTestIds } from './AnnotationPopover.tsx';
-import { appTestIds } from './App.tsx';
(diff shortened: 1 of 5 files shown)


jcarver989 (approved):
Thanks!

PR #192, opened 2026-05-20
feat: Keyboard shortcuts for sidebar

jcarver989:
## Summary

This PR makes PlanBridge comments work more like Google Docs sidebar comments and is prep work to support reviewing plan diff iterations. 

Would recommend reviewing commit by commit. 

## Changes

b358ec8 chore: Add react-hotkeys-hook
eb505ad refactor(annotation): Add useCommentNavigation and update other hooks to work with new sidebar
4492184 refactor(annotation): Update types and data model to support sidebar comments
e13c976 refactor(annotation): Update react componetns to support sidebar comments
80e0ee3 refactor: Extract subcomponents from App and update App to use new components


packages/annotation/src/useAnnotationInteractions.ts
```diff
@@ -160,80 +225,43 @@ export function useAnnotationInteractions({
...
+      openCurrent: openCurrentThreadComment,
+      previous: () => navigateThread('previous'),
+      total: navigableAnnotationCount,
+    },
+    resolvedThreads,
     setAnnotationHover,
   };
 }
 
-function computeResolvedAnnotations(
```
jcarver989, line 183: These move into another file as this hook was getting huge.

packages/annotation/src/useCommentNavigation.ts
```diff
@@ -0,0 +1,145 @@
...
+  activePosition: number;
+  currentThread: ResolvedAnnotationThread | null;
+  currentAnnotationId: string | null;
+  navigateThread: (direction: CommentNavigationDirection) => void;
+  navigableThreads: ResolvedAnnotationThread[];
+  openCurrentThreadComment: () => void;
+  total: number;
+}
+
+export function useCommentNavigation({
```
jcarver989, line 27: New hook to control navigating comments in the sidebar incl. keyboard shortcuts.

packages/annotation/src/useCommentNavigation.ts
```diff
@@ -0,0 +1,145 @@
+import { useCallback } from 'react';
+import { useHotkeys } from 'react-hotkeys-hook';
```
jcarver989, line 1: We're using this lib for keyboard shortcut bindings.

packages/annotation/src/annotationStyles.css
```diff
@@ -12,3 +12,25 @@
   background-color: color-mix(in oklch, var(--color-amber-300) 85%, transparent);
   color: var(--color-neutral-950);
 }
+
+.cb-draft-thread-attention {
+  animation: cb-draft-thread-attention 420ms ease-out 1;
+}
+
+@keyframes cb-draft-thread-attention {
```
jcarver989, line 20: Little animation for the comment card

packages/annotation/src/AnnotationDraftCommentComposer.tsx
```diff
@@ -0,0 +1,89 @@
...
+};
+
+export interface AnnotationDraftCommentComposerProps {
+  draft: ActiveCommentDraft;
+  onBodyChange: (body: string) => void;
+  onCancel: () => void;
+  onSave: () => void;
+}
+
+export function AnnotationDraftCommentComposer({
```
jcarver989, line 21: Shows up in the sidebar.

packages/annotation/src/useAnnotationInteractions.ts
```diff
@@ -97,17 +109,53 @@ export function useAnnotationInteractions({
     },
   });
 
+  const activateThread = useCallback((thread: ResolvedAnnotationThread) => {
```
blimmer: I think we don't need `useCallback` here (and in a few other places in this changeset). It'd be good for a review agent to re-scan the code, asking for adherence to modern React best-practices, specifically as it relates to React Compiler.

packages/annotation/src/annotationStyles.css
```diff
@@ -12,3 +12,25 @@
...
+  0% {
+    opacity: 0.65;
+    transform: translateY(0.5rem);
+  }
+
+  100% {
+    opacity: 1;
+    transform: translateX(0);
```
blimmer: Did you intend for this be to `translateY(0)`?

The animation looks a bit jumpy to me

packages/annotation/src/App.tsx
```diff
@@ -254,3 +210,128 @@ function resolveDocumentTitle(payload: AnnotationPayload | null): string {
...
+}
+
+function DiscardDraftDialog({ onConfirm, onOpenChange, open }: DiscardDraftDialogProps) {
+  return (
+    <AlertDialog onOpenChange={onOpenChange} open={open}>
+      <AlertDialogContent
+        data-testid={appTestIds.discardDraftDialog}
+        onEscapeKeyDown={(event) => {
+          event.preventDefault();
+          onConfirm();
```
blimmer: I think I might have asked for this 🙈 , but pressing escape twice now discards the draft. Should we assume that `esc` should do the "cancel" (keep) behavior instead?

blimmer (approved):
I'm at my stopping point for today. There's a ton here, added a few comments, but I'd say to merge this whenever you're ready. I tested it (with vimium disabled) in storybook and it looks good to me.

packages/annotation/src/useAnnotationInteractions.ts
jcarver989: Ah right I forgot about React compiler... good catch

packages/annotation/src/App.tsx
blimmer: `RemoveCommentDialog` doesn't set this, which defaults to the "cancel" action - we should probably do that here for consistency in our dialogs

packages/annotation/src/useCommentNavigation.test.tsx
```diff
@@ -0,0 +1,262 @@
...
+      id: thread.id,
+      anchor: thread.subject.anchor,
+      range,
+      target: null,
+      unresolved,
+      quote: thread.subject.anchor.quote.exact,
+      comments: [{ kind: 'saved', threadId: thread.id, message: primaryMessage, isPrimary: true }],
+    };
+  },
+  buildDraft({ id }: { id: string }): ResolvedAnnotationThread {
```
blimmer: Should this be an upstream fishery factory?

blimmer (approved):
I'd like to fix the escape behavior. let's ship the rest

packages/annotation/src/useCommentNavigation.test.tsx
jcarver989: Yea good call.

PR #196, opened 2026-05-21
feat: scaffold contextbridge review subcommand and UI

blimmer:
## Summary

Stands up the surfaces and plumbing for the upcoming `contextbridge review` code-review flow: a new CLI subcommand registered in help (marked `[WIP]`), a `startReviewServer` route in `@contextbridge/server`, a new `@contextbridge/review` Vite+React package whose bundle the CLI embeds, and wiring to publish its placeholder story through the shared `@contextbridge/storybook` host introduced in #195. The diff viewer itself lands in a follow-up per `docs/superpowers/specs/2026-05-20-review-feature-design.md`.

Also extracts the env-parsing portion of the build-define logic into `parseBuildEnv` + `cbBuildDefinesFromEnv` helpers in `@contextbridge/context/build` and threads them through every browser-UI `vite.config.ts` — behavior-preserving, but the diff touches each package's vite config.

## Review focus

- **`[WIP]` flagging in CLI help.** `contextbridge review` is visible in `contextbridge --help` once installed but is intentionally non-functional. Both the commander `summary` and `description` carry `[WIP]`, and the stdout message on Ctrl+C also carries it. Worth a read on whether the user-visible copy lands the right tone — clear enough that an early adopter won't expect feedback capture, but still describes the intended surface forward-looking.
- **`cbBuildDefinesFromEnv()` refactor.** Pure behavior-preserving extraction — every browser-UI `vite.config.ts` drops ~10 lines of env-parsing in favor of one helper call. Tightly coupled to the storybook-extraction merge (#195) since the annotation `vite.config.ts` was rebased through the conflict. Worth confirming the helper covers every env permutation the inline code did.

## Commits

History was rebased into 5 chunks so reviewers can read by domain. Each chunk is not guaranteed to build on its own (the helpers used in `packages/review/vite.config.ts` land in chunk 4); the final state on the branch tip is green.

- [`4367911`](https://github.com/contextbridge/planbridge/commit/436791171f0631d7d1293fa8b274a9b6088f518d) — scaffold `@contextbridge/review` Vite+React placeholder app (also aligns react/react-dom pins and declares them as direct deps)
- [`0a7b066`](https://github.com/contextbridge/planbridge/commit/0a7b06676f450fd8eb01419e0de031d7f494e6df) — add `startReviewServer` in `@contextbridge/server` (GET / serves the bundle)
[...]

PR #198, opened 2026-05-21
chore(deps): centralize shared deps via Bun workspace catalog

blimmer:
## Summary

Hoists 22 shared dependencies into a `workspaces.catalog` block in the root `package.json`; each consumer now references them as `"<dep>": "catalog:"` so version drift across packages becomes structurally impossible. Drifted ranges were bumped to npm latest in the same step (notably `lucide-react` v0 → v1, unifying `ui`/`annotation` with `website`), and the manual playwright drift check in CI collapses to a single root-catalog read since parity is now guaranteed.

This replaces the hacky cross-package `jq` compare flagged in #196 ([discussion_r3284179604](https://github.com/contextbridge/planbridge/pull/196#discussion_r3284179604)).

## Review focus

The `lucide-react` v0 → v1 bump is the only semantic change worth a second look — `just verify` is green (icons used in `ui`/`annotation` are `Check`/`ChevronRight`/`Circle`/`CircleHelp`/`MessageSquare`/`X`/`Sparkles`, none are brand icons removed in v1), but it's a major version jump bundled into a deps-refactor PR. Everything else is mechanical version unification at the caret-range floor.

## Commits

- [`0d24e70`](https://github.com/contextbridge/planbridge/commit/0d24e704aaa66753a1b124c167cac5cd124bcb01) — chore(deps): centralize shared deps via Bun workspace catalog


jcarver989 (approved):
Thanks for cleaning up!

PR #215, opened 2026-06-09
feat: add element annotations with a Mermaid diagram adapter

blimmer:
## Summary

Adds support for annotating rendered, non-text content in the annotation UI, with Mermaid diagrams as the first supported kind. The engine gains a kind-agnostic `ElementAdapter` layer: annotations can now be anchored to a diagram node, an edge, or a whole diagram by stable id, and the rest of the engine (resolver, CLI serializer, sidebar) stays generic over a new `element` anchor variant — adding a future content type needs no changes outside its adapter.

## Review focus

- **The `ElementAdapter` boundary** (`packages/annotation/src/element/ElementAdapter.ts` + `elementBlock.ts`). The contract is deliberately minimal: adapters own their DOM, sub-element tagging, anchor build/resolve, marker CSS, and hover; the only shared surface is the block-level data-attrs and a "rendered" event. Worth a hard look at whether that boundary is in the right place, or whether anything Mermaid-specific has leaked into the generic layer.
- **Anchor schema as a discriminated union** (`packages/shared/src/annotationSchema.ts`). `contentType` is a free string resolved against an adapter registry rather than an enum, so the schema never changes when a content type is added. The trade-off is the schema no longer constrains valid content types — that's enforced at the registry instead.

## Commits

- [`5e4a18c`](https://github.com/contextbridge/planbridge/commit/5e4a18cd1fbec42e79824c15ddf89b0a3bc14fda) — feat(shared): model annotation anchors as a text|element union (the data model)
- [`ba83a98`](https://github.com/contextbridge/planbridge/commit/ba83a98702c25dabcffbc6f4b2e5f60ac40f3d99) — feat(cli): describe annotation focus generically across anchor kinds (agent-facing serialization)
- [`125e8a0`](https://github.com/contextbridge/planbridge/commit/125e8a04b999e8f890aa5516202281c5736396f1) — feat(annotation): add generic element-annotation framework (the kind-agnostic engine, no concrete adapter registered yet)
- [`019a8f8`](https://github.com/contextbridge/planbridge/commit/019a8f8526fcaff7808e1d6a0f52dbf3a5da1c92) — feat(annotation): add Mermaid diagram adapter (first concrete adapter + e2e tests through the real `mermaid` dep)

> Reviewable commit-by-commit, bottom-up: each commit depends only on the ones before it, and the generic framework lands with an empty adapter registry so the abstraction can be judged before the ~700-line Mermaid implementation.

### DIFF
--- a/packages/annotation/src/AnnotatedMarkdown.tsx
+++ b/packages/annotation/src/AnnotatedMarkdown.tsx
@@ -1,10 +1,17 @@
 import type { AnnotationTargetKind, Asset } from '@contextbridge/shared/annotationSchema';
 import { cn } from '@contextbridge/ui/lib/utils';
+import type { Element } from 'hast';
 import { createElement } from 'react';
 import type { ComponentPropsWithoutRef, ComponentType, JSX, Ref } from 'react';
 import ReactMarkdown, { type ExtraProps } from 'react-markdown';
 import rehypeHighlight from 'rehype-highlight';
 import remarkGfm from 'remark-gfm';
+import { adapterForLang } from './element/ElementAdapter.ts';
+import {
+  ELEMENT_LANG_PROPERTY,
+  ELEMENT_SOURCE_PROPERTY,
+  rehypeElementSources,
+} from './element/rehypeElementSources.ts';
 
 export const annotatedMarkdownTestIds = {
   container: 'plan-review-markdown-plan',
@@ -40,7 +47,9 @@ export function AnnotatedMarkdown({ content, containerRef, onMouseUp, assets }: 
     >
       <ReactMarkdown
         remarkPlugins={[remarkGfm]}
-        rehypePlugins={[[rehypeHighlight, { detect: false, ignoreMissing: true }]]}
+        // rehypeElementSources must run before rehypeHighlight so a claimed block keeps its raw
+        // source instead of being tokenized into highlight spans.
+        rehypePlugins={[rehypeElementSources, [rehypeHighlight, { detect: false, ignoreMissing: true }]]}
         components={components}
       >
         {content}
@@ -146,7 +155,28 @@ const markdownComponents = {
     targetKind: 'block',
     className: 'border-l-4 border-chart-3/60 bg-chart-3/10 px-5 py-3 text-foreground/85 italic',
   }),
-  pre: AnnotatablePre,
+  pre: ({ node, children, ...props }: AnnotatableProps<'pre'>) => {
+    const fallback = (
+      <AnnotatablePre node={node} {...props}>
+        {children}
+      </AnnotatablePre>
+    );
+    const block = elementBlockFromPre(node);
+    const adapter = block ? adapterForLang(block.lang) : undefined;
+    if (!block || !adapter) {
+      return fallback;
+    }
+    const Block = adapter.Block;
+    return (
+      <Block
+        source={block.source}
+        contentType={adapter.contentType}
+        startLine={node?.position?.start.line}
+        endLine={node?.position?.end.line}
+        fallback={fallback}
+      />
+    );
+  },
   tr: annotatable('tr', {
     targetKey: 'tr',
     targetKind: 'table-row',
@@ -181,3 +211,12 @@ const markdownComponents = {
     <AnnotatableAnchor {...rest} rel={rel ?? 'noreferrer'} target={target ?? '_blank'} />
   ),
 };
+
+// Reads the source + language an element adapter stashed on a fenced code block (see
+// rehypeElementSources). Returns undefined for ordinary code blocks, which render normally.
+function elementBlockFromPre(node: Element | undefined): { lang: string; source: string } | undefined {
+  const code = node?.children.find((child): child is Element => child.type === 'element' && child.tagName === 'code');
+  const source = code?.properties[ELEMENT_SOURCE_PROPERTY];
+  const lang = code?.properties[ELEMENT_LANG_PROPERTY];
+  return typeof source === 'string' && typeof lang === 'string' ? { lang, source } : undefined;
+}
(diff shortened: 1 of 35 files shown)


packages/shared/src/annotationSchema.ts
```diff
@@ -72,6 +78,39 @@ export const StoredAnnotationAnchorSchema = z.object({
...
+export type TextAnnotationAnchor = z.infer<typeof TextAnnotationAnchorSchema>;
+
+/**
+ * An annotation anchored to a rendered, non-text element (a diagram node, a chart bar, …),
+ * located by stable id within its block — unlike {@link TextAnnotationAnchorSchema}, which is
+ * re-located by text quote/position. Produced and consumed by the matching ElementAdapter in
+ * `@contextbridge/annotation`; the rest of the engine (resolver, CLI, sidebar) treats it
+ * generically via these fields, so a new content type needs no changes outside its adapter.
+ */
+export const ElementAnnotationAnchorSchema = z.object({
```
blimmer, line 90: This covers annotating arbitrary elements (like mermaid components)

packages/cli/src/formatters/annotation/markdown.ts
```diff
@@ -100,6 +101,19 @@ function sliceSource(contentLines: string[], { start, end }: SourceLineRange): s
...
 
+function renderAnnotationFocus(anchor: StoredAnnotationAnchor): string | undefined {
+  if (anchor.kind === 'text') {
+    const highlighted = formatHighlighted(anchor.quote.exact);
+    return highlighted ? `the highlighted text: ${highlighted}` : undefined;
+  }
+
+  // Element anchors carry the agent-facing wording the adapter chose at capture time
+  // (descriptor + label), so this stays generic across content types: a whole-block
+  // annotation (no element id) describes itself by descriptor alone.
```
blimmer, line 112: I kinda liked having this inverted so the adapter knows how to describe its highlighted elements.

My goal is to support other rich context without having to touch a million files.

packages/annotation/src/element/ElementAdapter.ts
```diff
@@ -0,0 +1,52 @@
...
+ * this interface and register the instance in {@link elementAdapters} to make a kind of rendered
+ * content annotatable; the resolver, interaction hook, CLI serializer, and sidebar are generic
+ * over `element` anchors and need no per-type changes.
+ *
+ * An adapter is self-contained: it renders its own DOM, tags its own sub-elements (private
+ * data-attributes), builds and resolves anchors against those tags, and owns its marker and hover
+ * CSS. The only shared surface is the block-level contract in `elementBlock.ts`, which lets the
+ * interaction hook route events to the owning adapter.
+ */
+export interface ElementAdapter {
```
blimmer, line 18: Here's the common "rich media" adapter

packages/annotation/src/AnnotatedMarkdown.tsx
```diff
@@ -146,7 +155,28 @@ const markdownComponents = {
     targetKind: 'block',
     className: 'border-l-4 border-chart-3/60 bg-chart-3/10 px-5 py-3 text-foreground/85 italic',
   }),
-  pre: AnnotatablePre,
+  pre: ({ node, children, ...props }: AnnotatableProps<'pre'>) => {
```
blimmer, line 158: I assume, for now, that most of these extensions would be in code fences, but the schema should be adaptable.

packages/annotation/src/annotationResolvers.ts
```diff
@@ -98,5 +140,32 @@ function compareResolvedThreads(left: ResolvedAnnotationThread, right: ResolvedA
...
+  }
+
+  return secondaryOrderKey(left.anchor) - secondaryOrderKey(right.anchor);
+}
+
+/**
+ * Stable tiebreaker for two threads that share a start line (e.g. comments on different nodes of the
+ * same diagram). Text anchors order by character offset; element anchors have none, so hash their id.
+ */
+function secondaryOrderKey(anchor: StoredAnnotationAnchor): number {
```
blimmer, line 157: Might be a bit belt-and-suspenders, but felt OK.

packages/annotation/src/App.mermaid.test.tsx
```diff
@@ -0,0 +1,235 @@
...
+import { elementBlockAttrs } from './element/elementBlock.ts';
+import { mermaidAttrs } from './element/mermaid/MermaidBlock.tsx';
+import { annotateAndSubmit, renderApp, saveAnnotation } from './testHelpers/index.tsx';
+
+// End-to-end through the real `mermaid` dependency: a plan with a fenced mermaid block flows through
+// rehypeElementSources → AnnotatedMarkdown → MermaidBlock (renders the SVG and tags its nodes/edges)
+// → useElementTargets → the mermaid adapter's buildAnchor → the submitted payload. Rendering with
+// the actual library — not a hand-built SVG — is the point: a mermaid upgrade that changes the
+// rendered structure the adapter keys off (node element ids, edge-label markup) breaks these tests.
+describe('App — Mermaid diagram annotation', () => {
```
blimmer, line 17: This should be a good integration test to catch things that break in the mermaid library / change between versions.

packages/annotation/src/element/rehypeElementSources.ts
```diff
@@ -0,0 +1,48 @@
...
+export const ELEMENT_LANG_PROPERTY = 'dataElementLang';
+
+const LANGUAGE_PREFIX = 'language-';
+
+// Runs before rehype-highlight: for any fenced code block whose language a registered adapter
+// claims, stashes the raw source + language onto the <code> node and drops the language class so
+// the highlighter leaves it alone. Reading raw text here (pre-highlight) avoids reconstructing it
+// from tokenized highlight spans later. Registry-driven, so adding a content type needs no change
+// here.
+export function rehypeElementSources() {
```
jcarver989, line 15: This function appears to strip language names to avoid syntax highlighting so a better method name might be something like `stripLanguageClasses` or similar.

jcarver989 (approved):
Overall changes seem good. Although a lot more complex than I was hoping for this feature.

PR #218, opened 2026-06-10
refactor(annotation): simplify mermaid element pipeline

blimmer:
## Summary

Replaces the bespoke `rehypeElementSources` rehype plugin with rehype-highlight's built-in `plainText` option, then standardizes the element-adapter API naming. The custom plugin existed only to capture a fenced block's raw source before tokenization and strip its language class so the highlighter would skip it — `plainText` does both directly, leaving listed languages as untouched raw text with their `language-*` class intact. Net result: ~100 fewer lines, one fewer direct dependency (`unist-util-visit` out, `hast-util-to-string` in), and no behavior change.

## Review focus

The pipeline now leans on rehype-highlight's `plainText` semantics rather than ordering a custom plugin before the highlighter. Worth confirming `plainText` keeps the `language-*` class on skipped blocks (the adapter routing depends on it) and that reading raw source via `hast-util-to-string` off the hast node is equivalent to what the deleted plugin smuggled through a data-property. Routing is no longer unit-tested directly — it's covered end-to-end by `App.mermaid.test.tsx` (a mermaid block only renders if routing resolves) and `App.test.tsx` (ordinary code blocks still highlight); the now-redundant `adapterForLang` test was dropped.

## Commits

- [`feab94a`](https://github.com/contextbridge/planbridge/commit/feab94a45cd57dbcdaca51f54a987d9ec99950a9) — render element blocks via rehype-highlight plainText (drop the custom plugin; `ElementAdapter` declares `languages: string[]` instead of a `claims(lang)` predicate)
- [`b72fc26`](https://github.com/contextbridge/planbridge/commit/b72fc26c62e15905721cc8a723957b2e1838f324) — unify element-adapter export names on the `elementAdapter` prefix and the full word "Language"


jcarver989 (approved):
Thanks nice cleanup to use the plugin.

Issue #224, opened 2026-06-17
bug: severe per-keystroke typing lag in the annotation comment box (Firefox) when a large selection is highlighted


blimmer:
### What happened?

Typing into an annotation comment is roughly **1–2 s per character** when the active draft covers a **large text selection** (a full multi-line paragraph, or several). It is **Firefox-specific** — Chrome stays responsive. The root cause is that the CSS Custom Highlight registry is torn down and rebuilt on **every keystroke**, re-registering the entire draft-range highlight even though the highlighted ranges never change. Small selections feel fine, which is why it isn't always obvious.

### contextbridge version

`0.9.1-alpha.2` (where it was first observed). Reproduced and profiled against current `main` built from source (reports `0.0.0-development`); the responsible code is unchanged on `main`.

### Operating system

macOS

### Steps to reproduce

1. Open a large markdown document in the browser UI: `contextbridge open big.md`. Profiled with a generated ~117 KB / 1,425-line lorem-ipsum document with one mermaid diagram — **no proprietary content is needed to reproduce**.
2. In **Firefox**, drag-select a large span of text — a full multi-line paragraph or several — to open a new comment draft. The selection is highlighted (amber).
3. Type into the comment textarea.

### What did you expect to happen?

Typing in the comment box should be instant regardless of how large the highlighted selection (or the document) is. The comment text is independent of the document highlights, so editing it should not touch the highlight pipeline at all.

### What actually happened?

~100 ms+ per keystroke in Firefox, compounding into seconds of perceived lag; fine with a small selection. Every keystroke re-runs the highlight-sync effect, which deletes all highlight buckets and re-registers `new Highlight(draftRange)` over the **entire** selection — even though no range changed. The over-firing is an identity-churn artifact:

- `useAnnotationState.ts:77` — `setDraftBody` → `setActiveDraft((c) => ({ ...c, body }))` produces a new `activeDraft` object identity on every keystroke.
- `useAnnotationInteractions.ts:56` — `resolveAnnotationThreads(textIndex, threads, activeDraft)` returns a brand-new `resolvedThreads` array (the resolved ranges are identical; only the array identity changed).
- `useAnnotationInteractions.ts:155-165` — the highlight-sync `useEffect` has dependency `[draftRange, highlightedAnnotationId, resolvedThreads]`, so the new array re-runs it every keystroke.
[...]

PR #225, opened 2026-06-17
fix: stop draft typing from resyncing highlights

blimmer:
## Summary

Fixes the Firefox-specific annotation composer lag from #224 by keeping the live draft comment body out of the draft descriptor that feeds annotation range resolution and CSS Custom Highlight sync. Typing in the sidebar now updates only the composer body, while the highlighted document range remains stable.

This also clears stale draft body state when an edited thread is removed and adds a browser regression test proving textarea input no longer calls `CSS.highlights.set/delete`.

## Review focus

Please focus on the React state split between the draft descriptor and draft body: the goal is to keep the descriptor identity stable for highlight/marker effects while still preserving existing dirty-draft and save behavior.

## Commits

- [`52a426d`](https://github.com/contextbridge/planbridge/commit/52a426d) — split live draft body from the annotation descriptor to avoid highlight churn
- [`fc2f4ee`](https://github.com/contextbridge/planbridge/commit/fc2f4ee) — add regression coverage, clear stale draft body state, and remove local-only lint ignore

### DIFF
--- a/packages/annotation/src/annotationTypes.ts
+++ b/packages/annotation/src/annotationTypes.ts
@@ -40,18 +40,20 @@ export interface SelectableTextIndex {
   resolveTarget(targetId: string): AnnotatableTarget | null;
 }
 
+// The live draft body is tracked separately in useAnnotationState — deliberately not on this
+// descriptor. Editing the body must not change this object's identity, because it flows through
+// resolveAnnotationThreads into the CSS Custom Highlight + element-marker effects; a per-keystroke
+// identity churn there rebuilds the whole highlight range and is pathologically slow in Firefox.
 export type ActiveCommentDraft =
   | {
       kind: 'new-thread';
       anchor: StoredAnnotationAnchor;
-      body: string;
     }
   | {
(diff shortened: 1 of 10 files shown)


packages/annotation/src/annotationTypes.ts
```diff
@@ -40,18 +40,20 @@ export interface SelectableTextIndex {
   resolveTarget(targetId: string): AnnotatableTarget | null;
 }
 
+// The live draft body is tracked separately in useAnnotationState — deliberately not on this
+// descriptor. Editing the body must not change this object's identity, because it flows through
+// resolveAnnotationThreads into the CSS Custom Highlight + element-marker effects; a per-keystroke
+// identity churn there rebuilds the whole highlight range and is pathologically slow in Firefox.
```
blimmer, line 46: This was the crux of the problem. We invalidated the anchor each time the body changed, which caused the poor performance. Now it's tracked with two separate properties to avoid that.

jcarver989 (approved):
Thanks for the fix!

PR #239, opened 2026-07-09
fix: order mermaid diagram comments by document position

### LINKED ISSUE #238: bug: comments on mermaid diagrams always sink to the bottom of the comments sidebar
### What happened?

When reviewing a plan that contains a mermaid diagram, comments attached to the diagram (a node, an edge, or the diagram as a whole) always show up at the **bottom** of the comments sidebar, below every text-selection comment — even when the diagram is at the very top of the document.

All other comments are listed in document order, so the sidebar reads correctly until a diagram comment is involved. A comment on a diagram in the first section renders after a comment on prose in the last section, which makes the review hard to scan top-to-bottom.

### contextbridge version

0.9.2-alpha.1

### Operating system

macOS

### Steps to reproduce

1. Run `contextbridge plan` with a plan whose markdown starts with a ```mermaid diagram, followed by several paragraphs of prose.
2. In the review UI, add a comment to a node in the diagram.
3. Add a comment to a text selection in a paragraph further down the document.
4. Look at the order of the two comments in the sidebar.

Comments listed in document order: the diagram comment first (the diagram is at the top of the document), the prose comment after it.

### What actually happened?

The diagram comment is listed last, below all text-selection comments, regardless of where the diagram appears in the document.

### Browser and version

Reproduced in Chrome 137; not browser-specific.

### Anything else

A related symptom, likely the same underlying issue: diagram comments are also skipped by the sidebar's prev/next comment navigation and are not counted in the "Comment X of Y" indicator — only text-selection comments participate.

blimmer:
## Summary

Comments anchored to mermaid diagram elements always sank to the bottom of the comments sidebar and were skipped by prev/next navigation and the "Comment X of Y" counter, because `thread.range` doubled as the ordering and navigation-eligibility signal — and element anchors never resolve to a DOM Range by design. Sidebar ordering now uses the stored anchor's source lines (the one position every anchor kind shares), and navigation eligibility keys off the semantic `unresolved` field. Closes #238.

## Review focus

- Ordering now comes from creation-time source coordinates rather than the live DOM. The two diverge only for pre-existing threads whose anchors get quote-relocated after content drift — those sort at their original location. Unresolved ("Needs restore") threads also now sit at their last-known position instead of clumping at the end.
- Diagram comments becoming navigable is a deliberate behavior change: the "Comment X of Y" total grows wherever element threads are present.

## Commits

- [`89b4452`](https://github.com/contextbridge/planbridge/commit/89b44523aa4f20c87410276bb8b4b598a58cd2e4) — fix: order mermaid diagram comments by document position

### DIFF
--- a/packages/annotation/src/App.mermaid.test.tsx
+++ b/packages/annotation/src/App.mermaid.test.tsx
@@ -6,10 +6,12 @@ import { afterEach, describe, expect, it } from 'vitest';
 import { userEvent as browserUserEvent } from 'vitest/browser';
 import { annotatedMarkdownTestIds } from './AnnotatedMarkdown.tsx';
 import { annotationDraftCommentComposerTestIds } from './AnnotationDraftCommentComposer.tsx';
+import { annotationThreadCardTestIds } from './AnnotationThreadCard.tsx';
 import { appTestIds } from './App.tsx';
+import { commentsSidebarTestIds } from './CommentsSidebar.tsx';
 import { elementBlockAttrs } from './element/elementBlock.ts';
 import { mermaidAttrs } from './element/mermaid/MermaidBlock.tsx';
-import { annotateAndSubmit, renderApp, saveAnnotation } from './testHelpers/index.tsx';
+import { annotateAndSubmit, drag, renderApp, saveAnnotation } from './testHelpers/index.tsx';
 
 // End-to-end through the real `mermaid` dependency: a plan with a fenced mermaid block flows through
 // AnnotatedMarkdown → MermaidBlock (renders the SVG and tags its nodes/edges)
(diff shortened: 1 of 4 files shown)


packages/annotation/src/annotationResolvers.ts
```diff
@@ -124,30 +124,12 @@ function withActiveDraftThread(
   ];
 }
 
+// Source lines are the one position every anchor kind shares — element anchors never resolve to a DOM Range.
```
blimmer, line 127: This actually get a bit less complicated with this fix

packages/annotation/src/App.mermaid.test.tsx
```diff
@@ -165,6 +167,32 @@ describe('App — Mermaid diagram annotation', () => {
     expect(annotated.hovered).toBe(annotated.resting);
   });
 
+  it('orders the sidebar by document position when a diagram comment precedes a text comment', async () => {
```
blimmer, line 170: I TDD'ed this

jcarver989 (approved):

PR #246, opened 2026-07-22
refactor(annotation): use #src/* subpath imports for cross-directory refs

ghost:
## What & why

The repo's import convention (AGENTS.md → Conventions → Imports) is unambiguous:

> **Subpath imports** for intra-package refs: every package declares `"imports": { "#src/*": "./src/*" }`... Code writes `import { x } from '#src/context.ts';` for cross-directory imports. Same-directory siblings may stay relative.

This is the **dominant, established pattern** — 148 `#src/*` imports across the packages that declare the field. The **`@contextbridge/annotation` package was the sole holdout**: it never declared the `#src/*` alias in its `package.json` and used relative parent imports (`from '../...'`) for every cross-directory reference. That made it the single most egregious, self-contained violation class in the codebase.

## Changes

- **`packages/annotation/package.json`** — added the missing `"imports": { "#src/*": "./src/*" }` declaration (it was the only src-bearing package missing it, aside from the `review` scaffold which has no cross-directory imports).
- **Rewrote all 26 `../` cross-directory imports to `#src/*`** across 11 source/test files (`demo/`, `element/`, `element/mermaid/`, `testHelpers/`). Same-directory sibling imports (`./x`) were intentionally left relative, per the convention. Import identifiers, `type` modifiers, and ordering were preserved exactly — this is a pure path rewrite with no runtime/behavior change.

## Scope notes

- The only remaining `from '../...'` imports in the repo are `with { type: 'text' }` build-time asset embeds that reference files **outside** their package (`../../../annotation/dist/index.html`, `../../../harnessIntegrations/.../SKILL.md`). These are not `src/`-relative and cannot be expressed as `#src/*`, so they are correctly out of scope.

## Verification

- `bun run typecheck` (annotation) — passes
- `bun run lint` (repo, `--max-warnings 0`) — passes
- `bun run format:check` — passes
- Vite production build (annotation) — passes, 4730 modules transformed (confirms Vite resolves the `#src/*` subpath alias, not just `tsc`)
- Full vitest browser suite (annotation) — **159 tests / 13 files pass** (installed `chromium-headless-shell` locally to run browser-mode tests)

## Other violation classes noted for future runs (intentionally not touched here — one class per run)

1. **Bun-native APIs** — several Bun-only production files still use `node:fs` `readFileSync`/`readdirSync` instead of `Bun.file`/`Bun.Glob` (e.g. `skills/src/skills.ts`, `skills/src/handlebars.ts:27` — ironically the rule's own in-repo example file, `instrumentation/src/node/*.ts`). Deferred because migrating sync reads to `Bun.file().text()` forces functions async and cascades to callers — a behavior-sensitive change that doesn't fit the "conservative, no behavior change" bar.
2. **Destructured defaults over `??`** — a handful of per-field `options.x ?? default` on options bags (e.g. `buildKeyEvent` in `useSubmitOnCmdEnter.test.ts`, `options.submission` in `testHelpers/annotationFakes.ts`). Low frequency; most `??` in the codebase are legitimate null-coalescing on lookups, not options-bag defaults.


blimmer:
ContextBot has [been retired](https://contextbridge.ai/contextbot/). Closing.

Issue #247, opened 2026-07-27
feat: add option to exit plan mode into auto mode in Claude Code


jpalharini:
### What's the problem?

Current hook for Claude Code hardcodes mode as `acceptEdits` (see [source](https://github.com/contextbridge/planbridge/blob/b49bc7c837ba11396e9fcdb9e69ce805393ff1c3/packages/cli/src/formatters/plan/claudeHookResponse.ts#L32)).

It'd be nice to have the option to approve and use auto mode, like it happens through CLI.

### Proposed solution

Ideally offer two buttons in the browser view, but a configuration option on `contextbridge install` would potentially be simpler.

jcarver989:
Hey @jpalharini  thanks for filing! 

I saw that you forked the repo and had started working on this. Is this something you'd like to take a crack at? If not, I'm also happy to open a PR to add this for you.

jpalharini:
@jcarver989 I can take a crack at it, sure. However, I'm relying heavily on Claude to do the work since I have almost no experience with TypeScript.

If you don't mind, I'll do a review pass tomorrow on the PR in my fork and open it here.

jcarver989:
Sure, take your time. I'll be traveling over the next few days (back on Tuesday) and happy to review any PRs you have open then!

jpalharini:
Opened #249 and marked ready for review.

PR #249, opened 2026-07-31
feat: let reviewers approve a plan into Claude Code auto mode

jpalharini:
## Summary

The `PermissionRequest:ExitPlanMode` hook always exits plan mode into `acceptEdits`, so a reviewer who wants Claude to keep going without prompts has to switch modes by hand right after approving.

This PR adds a mode selector next to "Approve Plan" in the browser UI with two options, Accept Edits (still the default) and Auto, and carries the choice through the submission payload into the hook's `setMode` permission update. The choice is deliberately not persisted: every review starts back at Accept Edits. Aims to address #247.

## Review focus

1. **Scope of the enum.** `ApprovalModeSchema` is `z.enum(['acceptEdits', 'auto'])`, not the SDK's full `PermissionMode`. `packages/shared` has no dependency edge on `@anthropic-ai/claude-agent-sdk`, so the subset relationship is held by a comment there plus the contextual typing of `updatedPermissions` in `claudeHookResponse.ts`, which fails to compile if the SDK ever drops `'auto'`. We may want to also add `bypassPermissions`, although that no longer seems to be an option in CLI.
2. **Where the mode is echoed to the reviewer.** The mode appears as a suffix on the primary button ("Approve Plan (Auto)") in two places: the submit bar and the close-review guard dialog, since that dialog also submits the selected mode. Both go through one predicate (`shouldSuffixApprovalMode`) so the two surfaces cannot disagree, and neither shows a suffix at the default. Worth a look at whether `App.tsx` reaching into `SubmitBar.tsx` for that predicate is the placement you want.
3. **`approvalMode: status === 'approved' ? approvalMode : 'acceptEdits'`** in `useAnnotationState.ts` keeps a rejected submission from asserting a mode nothing will honor. It is one ternary guarding a wire payload, and it is intentional rather than leftover.

`ApprovalModeSchema.default('acceptEdits')` makes the field required in the inferred output type, which is why the CLI formatter tests moved onto the `annotationSubmission` fishery factory instead of hand-rolled literals.

`just verify` is green locally, including the vitest browser-mode suites. Docs updated at `packages/website/src/content/docs/usage/claude-code.mdx`.

## Commits

- [`512af9d`](https://github.com/contextbridge/planbridge/commit/512af9d) feat(shared): carry an approval mode on plan submissions. Adds `ApprovalModeSchema`, the defaulted field on `AnnotationSubmissionSchema`, the factory default, and the test migration onto the factory.
- [`41337f6`](https://github.com/contextbridge/planbridge/commit/41337f6) feat: let reviewers approve a plan into Claude Code auto mode. The `setMode` change in the hook formatter, the submit-bar selector, the close-dialog label, and the docs.

### DIFF
--- a/packages/annotation/src/App.stories.tsx
+++ b/packages/annotation/src/App.stories.tsx
@@ -164,6 +164,28 @@ export const Default: Story = {
   decorators: [withAppContext({ submitAnnotation: delayedSubmit })],
 };
 
+export const ApprovalModeSelector: Story = {
+  args: {
+    source: 'hook_claude',
+    initialPayload: {
+      contentKind: 'plan',
+      content: samplePlan,
+      title: 'Refactor auth middleware',
+      metadata: { entrypoint: 'hook_claude' },
+    },
+  },
+  render: Default.render,
+  decorators: [withAppContext({ submitAnnotation: delayedSubmit })],
+  parameters: {
+    docs: {
+      description: {
+        story:
+          'Zero-feedback review from a Claude Code hook: the compound submit control (primary Approve Plan button + chevron trigger opening the approval-mode radio menu) is visible for visual review.',
+      },
+    },
+  },
+};
+
 export const SeededReview: Story = {
   args: {
     initialPayload: {
--- a/packages/annotation/src/App.test.tsx
+++ b/packages/annotation/src/App.test.tsx
@@ -18,6 +18,7 @@ import { submitBarTestIds } from './SubmitBar.tsx';
 import { drag, pressSubmitShortcut, renderApp } from './testHelpers/index.tsx';
 import { themePickerTestIds } from './ThemePicker.tsx';
 import { updateNoticeCardTestIds } from './UpdateNoticeCard.tsx';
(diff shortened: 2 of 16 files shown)


blimmer:
@jpalharini thank you so much for your contribution and my sincere apologies for taking so long to get back to you on this.

We've been holding off on introducing a settings panel for a while, but things like this feel like they belong in one!

There's now a "gear" icon in the top right, with this setting for Claude Code:

Screenshot 2026-08-11 at 13 53 50

v0.11.0-alpha.0 is out now with those changes. I want to use it for a day or so before releasing to stable (which I intend to do tomorrow afternoon if it looks good). You're welcome to try out the alpha in the meantime if you'd like it sooner (see https://plan.contextbridge.ai/install/#installer-options and use the alpha flag).

Again, thank you for the contribution and your patience!

jpalharini:
That is awesome @blimmer!

Having this be part of some config was an option I had thought about, and it definitely looks and feels better this way.

Congrats on the great work with PlanBridge and I'm glad I could contribute to it a bit.

I'll take this for a spin tomorrow.

blimmer:
Release v0.11.0 is live with this change. Note that `auto` is the new default!

PR #253, opened 2026-08-10
fix(annotation): preserve line breaks in saved comment bodies

### LINKED ISSUE #252: bug: multi-line comment bodies collapse to a single line in the saved (non-editing) thread card
### What happened?

In the annotation UI, a comment whose body contains newlines renders correctly while it is being edited (the textarea preserves the line breaks), but once saved, the thread card collapses every newline into a single space. A comment written as three lines displays as one run-on line.

The saved comment body is rendered as raw text inside a `<p>` with no whitespace handling, so HTML's default whitespace collapsing applies:

`packages/annotation/src/AnnotationThreadCard.tsx:72-78`

```tsx
<p
  className="mt-3 text-sm leading-6 text-foreground/80"
  data-testid={annotationThreadCardTestIds.comment(comment.message.id)}
  key={comment.message.id}
>
  {comment.message.body}
</p>
```

The class list has no `whitespace-pre-wrap` (the only place that class appears in the package is `AnnotatedMarkdown.tsx:49`, scoped to inline code). The stored body is intact — this is purely a display bug, and the submitted payload is unaffected.

### contextbridge version

0.10.0 (dev build from `main` at commit `8a469e3`)

### Operating system

macOS

### Steps to reproduce

1. Run `contextbridge plan` with any plan document.
2. Select a passage in the plan and add an annotation comment.
3. In the comment textarea, type a body with blank lines between them, e.g.:
   ```
   Foo

   Bar

   Baz
   ```
4. Click **Save**.
5. Observe the saved comment in the thread card in the sidebar.

The saved comment renders with the same line structure the author typed, matching what the editing textarea shows.

### What actually happened?

The saved card renders `Foo Bar Baz` on one line. Re-opening the comment for editing shows the original three-line body, confirming the newlines are preserved in state and only lost at render time.

### Anything else?

Suggested fix: add `whitespace-pre-wrap` to the saved-comment `<p>` in `AnnotationThreadCard.tsx`. Worth deciding at the same time whether the body should be rendered as markdown instead of plain text — comment bodies come from a textarea and users may reasonably type lists or code fences — but `whitespace-pre-wrap` is the minimal correct fix for the mismatch between the editing and display states.

Also worth checking whether any other saved-body render sites have the same gap as more comment surfaces land.

blimmer:
## Summary

A saved comment body was rendered as raw text in a `<p>` with no whitespace handling, so HTML collapsed every newline into a space and a three-line comment displayed as one run-on line. Reopening it for editing showed the original lines, so the text was always intact in state and only the display was wrong. Adds `whitespace-pre-wrap`, a regression test, and a story. Closes #252.

## Review focus

The vitest config change is the part I'd look at. There was no way to test this as things stood: vitest had no `@tailwindcss/vite` plugin, so a test importing `styles.css` got the theme layer but none of the generated utilities, and every utility class resolved to its unstyled default. I added the plugin. It only affects tests that import `styles.css` themselves, which today means `App.mermaid.test.tsx` (still passing) and the new file. Those tests now measure real Tailwind output instead of unstyled markup, which is what you want, but it does change what they were asserting.

I also skipped markdown rendering, which the issue floated. Comment bodies are plain text the agent reads verbatim, and rendering markdown in the card while the textarea still shows source would recreate the same edit-versus-display mismatch this PR fixes. Happy to file that separately if you disagree.

## Commits

- [`4159460`](https://github.com/contextbridge/planbridge/commit/415946036350dfc08545149ac3683ddee1cc18ea) — enable Tailwind in vitest browser runs so utility classes actually resolve
- [`655bdd1`](https://github.com/contextbridge/planbridge/commit/655bdd156af6b291fb53a8981910927a4b31da58) — add `whitespace-pre-wrap`, the height-comparison regression test, and the `MultiLineComment` story

### DIFF
--- a/packages/annotation/src/AnnotationThreadCard.test.tsx
+++ b/packages/annotation/src/AnnotationThreadCard.test.tsx
@@ -1,0 +1,45 @@
+import '@contextbridge/ui/styles.css';
+import { commentMessage } from '@contextbridge/shared/testFactories';
+import { cleanup, render, screen } from '@testing-library/react';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { AnnotationThreadCard, annotationThreadCardTestIds } from './AnnotationThreadCard.tsx';
+import { resolvedAnnotationThread } from './testFactories.ts';
+
+describe('AnnotationThreadCard', () => {
+  afterEach(() => {
+    cleanup();
+  });
+
+  it('renders a saved comment body across the lines the author typed', () => {
+    const singleLine = renderSavedComment({ id: 'msg_single', body: 'Foo' });
+    const multiLine = renderSavedComment({ id: 'msg_multi', body: 'Foo\n\nBar\n\nBaz' });
+
+    expect(multiLine.getBoundingClientRect().height).toBeGreaterThan(singleLine.getBoundingClientRect().height * 2);
+  });
+});
+
+function renderSavedComment({ id, body }: { id: string; body: string }): HTMLElement {
+  const message = commentMessage.build({ id, body });
+  const thread = resolvedAnnotationThread.build({
+    id: `thr_${id}`,
+    comments: [{ kind: 'saved', threadId: `thr_${id}`, message, isPrimary: true }],
+  });
+
+  render(
+    <AnnotationThreadCard
+      draftBody=""
+      isActive={false}
+      isCurrent={false}
+      onClick={vi.fn()}
+      onDraftBodyChange={vi.fn()}
+      onDraftCancel={vi.fn()}
+      onDraftSave={vi.fn()}
+      onHoverChange={vi.fn()}
+      onRequestRemove={vi.fn()}
+      submitted={false}
+      thread={thread}
+    />,
+  );
+
+  return screen.getByTestId(annotationThreadCardTestIds.comment(message.id));
+}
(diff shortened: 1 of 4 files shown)


packages/annotation/src/AnnotationThreadCard.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+describe('AnnotationThreadCard', () => {
+  afterEach(() => {
+    cleanup();
+  });
+
+  it('renders a saved comment body across the lines the author typed', () => {
+    const singleLine = renderSavedComment({ id: 'msg_single', body: 'Foo' });
+    const multiLine = renderSavedComment({ id: 'msg_multi', body: 'Foo\n\nBar\n\nBaz' });
+
+    expect(multiLine.getBoundingClientRect().height).toBeGreaterThan(singleLine.getBoundingClientRect().height * 2);
```
blimmer, line 17: Kinda clever test, IMO. Good bot.

packages/annotation/src/App.stories.tsx
```diff
@@ -178,6 +200,27 @@ export const SeededReview: Story = {
   decorators: [withAppContext({ submitAnnotation: delayedSubmit })],
 };
 
+export const MultiLineComment: Story = {
```
blimmer, line 203: We'll also get visual chromatic coverage.

Issue #251, opened 2026-08-10
feat: user settings file with versioned schema and per-harness namespaces


blimmer:
### What's the problem?

Preference requests are piling up with no place to live:

- #247 / #249: choose which permission mode Claude Code enters after plan approval
- #178: opt out of auto-implementation after approval
- The theme picker from #242 saves to localStorage, but the review server binds a fresh ephemeral port each run (`resolveListenPort` defaults to `0` in `packages/server/src/annotation.ts:61`) and localStorage is origin-scoped. Every review is a new origin, so the theme choice doesn't actually stick for default users.

Building each of these as a one-off means another widget on the review screen or another storage hack. We need one home for user preferences before we take any of them.

### Proposed solution

A sparse, versioned settings file at `$XDG_CONFIG_HOME/contextbridge/settings.json`, next to the existing `anonymous_id`. The CLI process reads it directly; the browser UI reads and writes it through the review server.

Schema lives in `@contextbridge/shared/settingsSchema.ts` with harness-scoped namespaces:

```jsonc
{
  "version": 1,
  "ui": { "theme": "github-dark" },
  "harnesses": {
    "claude": { "planApprovalMode": "auto" }
  }
}
```

- `ui` holds browser-surface preferences. `harnesses.<id>` holds settings that only mean something to one harness, with an explicit Zod schema per harness (not a `z.record`) and a `satisfies Partial<Record<HarnessId, ...>>` check against the registry.
- The file is sparse: we only write explicit choices, never defaults. Most users never get a file, and changing a default in a later release is not a migration.
- Writes are patch-merges into a fresh read of the raw file, written atomically (temp + rename). Unknown keys survive, and an older binary can't clobber a file written by a newer one.
- Integer schema `version` with forward-only pure migration functions in shared. Runtime parsing is lenient (bad leaf value falls back to its default with a warning); test parsing is strict.

Version safety is enforced in `bun run test`:

- A frozen fixture corpus (`v1.kitchen-sink.json`, `v1.minimal.json`, ...) committed forever. Every fixture must migrate to a valid current document, and the migrated output is snapshotted.
- A committed JSON Schema snapshot of the current `SettingsSchema`, so any shape change shows up in the PR diff.
- Guards that every historical version has a migration and fixtures, plus a `future-version.json` fixture proving a downgraded binary degrades to defaults without rewriting the file.

A breaking schema change can't merge unless the PR carries a migration, new fixtures, and snapshot updates.

[...]

PR #254, opened 2026-08-10
feat(settings): add versioned settings contracts

blimmer:
## Summary

First layer of the user settings file from #251: the shared contracts, and nothing else — no consumer wiring lands here. Adds a versioned settings schema (`@contextbridge/shared/settingsSchema`) with three strict shapes: a sparse `PersistedSettings` for the on-disk file, a `SettingsPatch` for browser → CLI writes, and a fully-defaulted `Settings` that consumers read. Also adds the `SettingsStore` interface, a shared theme enum that the annotation UI's catalog is now keyed by, and a `HARNESS_IDS` runtime array on `@contextbridge/harness`.

## Review focus

- Every schema is strict. A file with keys this binary doesn't know is not a valid settings document; there is no forward-compatibility looseness and no future-version handling. Multiple binary versions never share a machine (hooks shell out to `contextbridge` unversioned), so the only cross-version case is an old file meeting a newer binary, which a future v2 handles with a migration.
- The v1 fixtures in `settingsFixtures/` are the freeze: tests parse every fixture strictly, so any shape change fails the merge gate until it ships a migration story. No migration runner exists yet; machinery arrives when a v2 exists.

## Commits

- [`48517d9`](https://github.com/contextbridge/planbridge/commit/48517d94e0be3582af6aedbd39c279338c325b49) — feat(settings): add versioned settings contracts


packages/harness/src/types.ts
```diff
@@ -1,4 +1,5 @@
-export type HarnessId = 'claude' | 'codex' | 'gemini' | 'cursor' | 'aider' | 'opencode' | 'aether';
+export const HARNESS_IDS = ['claude', 'codex', 'gemini', 'cursor', 'aider', 'opencode', 'aether'] as const;
```
blimmer, line 1: So I can have harness-specific settings typed to these. I need the array so I can use it with zod.

packages/shared/src/settingsFixtures/v1.full.json
```diff
@@ -0,0 +1,6 @@
+{
```
blimmer, line 1: These are for future migration support.

packages/shared/src/settingsSchema.ts
```diff
@@ -0,0 +1,45 @@
+import type { HarnessId } from '@contextbridge/harness';
+import { z } from 'zod';
+import { ThemePreferenceSchema } from './themeSchema.ts';
+
+export const CURRENT_SETTINGS_VERSION = 1 as const;
```
blimmer, line 5: I have some logic for settings migrations, but the migration logic itself does not yet exist. We'd only need to add that if we need break the settings file, which I think we can avoid (ideally) for a while.

packages/shared/src/themeSchema.ts
```diff
@@ -0,0 +1,21 @@
...
+  'github-dark-default',
+  'ayu-dark',
+  'everforest-dark',
+  'dracula',
+  'catppuccin-mocha',
+  'gruvbox-dark-medium',
+  'nord',
+  'rose-pine',
+  'tokyo-night',
+] as const;
```
blimmer, line 18: So we don't duplicate between the settings and the UI/shiki logic.

packages/shared/src/settingsFixtures/v1.full.json
jcarver989: Hmm. I'm Not a big fan of these snapshot tests... for a few reasons:

1. I'm assuming we'd retain type-definitions / zod schemas around for some number of older schema versions. If that's true, I think it'd be nicer to have a Fishery factory for the legacy settings version(s) as that would be type-checked and thus guaranteed that it's a valid schema. Here, these captured snapshots have no compile time check that they're valid.   

2. Fixtures make tests harder to read vs a factory since it separates the data used in the test form the test itself, whereas the factory approach inlines the data into the test (so the reader doesn't have to jump around to different files). Generally, I'll only reach for fixtures if a factory isn't a good fit and the data is too verbose to read inside of a test (e.g. PatchWave's evals for the analysis report are verbose and aren't a good fit for a factory)

packages/shared/src/settingsSchema.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  CURRENT_SETTINGS_VERSION,
+  PersistedSettingsSchema,
+  SettingsPatchSchema,
+  resolveSettings,
+} from './settingsSchema.ts';
+
+const fixturesDirectory = resolve(import.meta.dir, 'settingsFixtures');
+
+describe('settings schema', () => {
+  it('parses and resolves every frozen v1 fixture', async () => {
```
jcarver989: Somewhat related to my other comment, but might consider using `it.each` along with a factory https://vitest.dev/api/test.html#test-each

packages/shared/src/settingsSchema.ts
```diff
@@ -0,0 +1,45 @@
...
+
+const uiSettingsShape = {
+  theme: ThemePreferenceSchema,
+} as const;
+const harnessSettingsShape = {} satisfies Partial<Record<HarnessId, z.ZodType>>;
```
jcarver989: Seems odd to leave the harness specific settings as `z.ZodType`, wouldn't we want each harness to have a strongly typed set of options?

packages/shared/src/settingsSchema.ts
```diff
@@ -0,0 +1,45 @@
+import type { HarnessId } from '@contextbridge/harness';
+import { z } from 'zod';
+import { ThemePreferenceSchema } from './themeSchema.ts';
+
+export const CURRENT_SETTINGS_VERSION = 1 as const;
+
+const uiSettingsShape = {
```
jcarver989: Not a big deal, but would personally drop the `Shape` suffix -- e.g. `const uiSettings =`. If you like it, I'm fine keeping it.

packages/shared/src/settingsSchema.ts
```diff
@@ -0,0 +1,45 @@
...
+  theme: ThemePreferenceSchema,
+} as const;
+const harnessSettingsShape = {} satisfies Partial<Record<HarnessId, z.ZodType>>;
+
+/**
+ * The on-disk settings document. Sparse: only keys the user explicitly set
+ * are written. Frozen once version 2 ships; migrate instead of editing.
+ */
```
jcarver989, line 15: Looks like the LLM is divulging it's future plans in this comment :)

packages/shared/src/settingsSchema.ts
```diff
@@ -0,0 +1,45 @@
...
+export const SettingsPatchSchema = z
+  .strictObject({
+    ui: z.strictObject(uiSettingsShape).partial(),
+    harnesses: z.strictObject(harnessSettingsShape),
+  })
+  .partial();
+export type SettingsPatch = z.infer<typeof SettingsPatchSchema>;
+
+/** Settings with every default applied. Consumers read this shape. */
+export const SettingsSchema = z.strictObject({
```
jcarver989, line 33: Cool. 

In general, I think it's a good idea to split the persisted versioned type from the internal in-memory type as it allows us to map the persisted type to a common internal type.

packages/shared/src/settingsSchema.ts
```diff
@@ -0,0 +1,45 @@
...
+  ui: z.strictObject(uiSettingsShape),
+  harnesses: z.strictObject(harnessSettingsShape),
+});
+export type Settings = z.infer<typeof SettingsSchema>;
+
+export function resolveSettings(persisted?: PersistedSettings): Settings {
```
jcarver989, line 39: Would call this `parseSettings` a you're parsing the on-disk settings into the internal format.

jcarver989 (approved):
Overall looks good, nothing blocking although per my comment I'm not a big fan of snapshot tests.

packages/shared/src/settingsSchema.ts
blimmer: Right - I actually reviewed this and left it as-is because when version 2 is introduced, we know that we need to migrate.

PR #255, opened 2026-08-10
feat(settings): persist settings through the local server

blimmer:
## Summary

Second layer of #251: persistence and plumbing. `SettingsStoreImpl` owns the sparse file at `$XDG_CONFIG_HOME/contextbridge/settings.json`, the CLI context carries a `SettingsStore`, and `runAnnotation` reads fresh settings per session into the frontend config — `FrontendConfig` gains its `settings` field here, alongside the code that fills it. The review server gains `POST /settings` for validated patches from the browser, and successful patches fold into what `GET /config` serves so a page load after a save sees the saved settings. Path resolution is shared with the anonymous-id store through a new `configDir` helper in `@contextbridge/instrumentation/node`.

## Review focus

- The store is deliberately plain: `Bun.file` read, `Bun.write` write, no atomic-rename or permission ceremony — matching how the anonymous-id store next door persists its file. The tradeoff is that a crash mid-write can tear the file; a torn file warns and reads as defaults, and refuses patches until removed.
- Patch conflict semantics: `read` never fails (a missing file is defaults silently; a malformed or invalid one logs a warning and resolves to defaults), but `patch` refuses with a `conflict` error rather than rewrite a file it can't parse as a valid v1 document.
- The file stays sparse: only explicit user choices are written, never defaults, so flipping a default in a later release is not a migration. A patch that changes nothing writes nothing.
- Also ports the fuller neverthrow rule (combinators over manual `isErr()` chains; `assert` narrowing over `_unsafeUnwrap` in tests) into `.claude/rules/`, which this PR's code and tests follow.

## Commits

- [`fb92c8d`](https://github.com/contextbridge/planbridge/commit/fb92c8d5516621d371ff146df6e88b376237a852) — feat(settings): persist settings through the local server

### DIFF
--- a/.claude/rules/neverthrow-error-handling.md
+++ b/.claude/rules/neverthrow-error-handling.md
@@ -1,0 +1,76 @@
+---
+paths:
+  - packages/**/*.ts
+  - packages/**/*.tsx
+---
+# Neverthrow Error Handling
+
+Fallible operations return `Result<T, E>` (from `neverthrow`) instead of throwing. Callers
+handle the error path explicitly. This extends the safe-wrapper rule in
+`.contextbridge/rules/error-handling-neverthrow.md`.
+
+## Prefer combinators over manual `isErr()` checks
+
+Chain the pipeline with combinators — `.andThen()` / `.map()` / `.mapErr()` / `.orElse()` /
+`.asyncAndThen()` — instead of awaiting each step and branching on `isErr()`. A long
+sequence of `if (x.isErr()) return err(x.error)` is a smell: it's the manual expansion of
+what `.andThen()` already does, and it grows the diff without adding logic. Let the error
+short-circuit through the chain; only the `Ok` path stays in your callbacks.
+
+```ts
+// Good — one chain, error short-circuits automatically
+private applyAndWrite(patch: SettingsPatch): ResultAsync<Settings, SettingsStoreError> {
+  return this.readPersisted().andThen((persisted) => {
+    const updated = applyPatch(persisted ?? emptyDocument(), patch);
+    return this.writePersisted(updated).map(() => resolveSettings(updated));
+  });
+}
+
+// Avoid — manual unwrap-and-rewrap at every step
+const readResult = await this.readPersisted();
+if (readResult.isErr()) return err(readResult.error);
+const updated = applyPatch(readResult.value ?? emptyDocument(), patch);
+const writeResult = await this.writePersisted(updated);
+if (writeResult.isErr()) return err(writeResult.error);
+// ...
+```
+
+Mechanics:
+
+- Return `ResultAsync<T, E>` directly from async fallible functions (it's awaitable to
+  `Result<T, E>`, so callers can still `await` + `isErr()`). Synchronous functions
+  return `Result<T, E>`.
+- Bridge sync→async with `.asyncAndThen()`; a `Result` from a parser or zod check flows
+  straight into an async chain without an intermediate `await`.
+- Wrap a throwing promise with `ResultAsync.fromPromise(promise, toError)`. Wrap a
+  `Promise<Result<…>>` (an already-fallible async API) with `new ResultAsync(promise)` so
+  it joins the chain without double-wrapping.
+- Inside an `.andThen()` callback you may return a plain `Result` (`ok(...)` / `err(...)`)
+  or another `ResultAsync` — mix freely. Short imperative guards (a validation `for` loop
+  returning `err(...)`) are fine *inside* a callback; the win is not re-checking the
+  surrounding pipeline by hand.
+- Use `.orElse()` for idempotent recovery — swallow expected conditions by returning
+  `ok(...)` so the operation is retry-safe. Use `.mapErr()` only to log/translate.
+- Keep try/catch + a `toError` normalizer for genuinely imperative code (pagination loops,
+  `for await`).
+
+`isErr()` propagation is still correct where a chain doesn't read better — but reach for it
+as the exception, not the default.
+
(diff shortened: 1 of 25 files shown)


.claude/rules/neverthrow-error-handling.md
```diff
@@ -0,0 +1,76 @@
+---
```
blimmer, line 1: I have another agent working on moving us to rulesync.

packages/cli/src/annotation/runAnnotation.test.ts
```diff
@@ -44,6 +44,20 @@ describe('runAnnotation', () => {
     expect(deps.port).toBe(3000);
   });
 
+  it('reads fresh settings for each session', async () => {
```
blimmer: review agent - remove this test. it's useless.

packages/cli/src/settings/SettingsStoreImpl.test.ts
```diff
@@ -0,0 +1,146 @@
...
+    expect(await store.read()).toEqual(resolveSettings());
+    expect(readFileSync(store.path, 'utf8')).toBe('{"version":');
+  });
+
+  it('reads an invalid settings document as defaults without rewriting it', async () => {
+    const store = createStore();
+    await writeSettings(store, '{"version":1,"ui":{"theme":"not-a-theme"}}');
+    expect(await store.read()).toEqual(resolveSettings());
+    expect(JSON.parse(readFileSync(store.path, 'utf8'))).toEqual({ version: 1, ui: { theme: 'not-a-theme' } });
+  });
```
blimmer: I suppose we could throw, too. Thoughts?

packages/instrumentation/src/node/anonymousId.ts
```diff
@@ -32,11 +28,3 @@ export function getOrCreateAnonymousId(env: AnonymousIdEnv): string {
   safeWrite(dir, path, id);
   return id;
 }
-
-function configDir(env: AnonymousIdEnv): string {
```
blimmer, line 36: hoisting

packages/cli/src/settings/SettingsStoreImpl.test.ts
```diff
@@ -0,0 +1,146 @@
...
+const temporaryDirectories: string[] = [];
+
+afterEach(async () => {
+  await Promise.all(
+    temporaryDirectories.splice(0).map(async (path) => {
+      await chmod(path, 0o700).catch(() => {});
+      await rm(path, { recursive: true, force: true });
+    }),
+  );
+});
```
jcarver989: For PatchWave evals I've been using `await using` to handle cleanup operations in tests (e.g. Docker containers, tmpdir removal etc) which I've really liked. Might consider using that here too.

packages/cli/src/settings/SettingsStoreImpl.test.ts
jcarver989: Hmm... yea I'd just make `parseSettings` (what you had as `resolveSettings` iirc) return an error `Result` if it's invalid.  Falling back to defaults can be confusing for users, e.g. "why is my theme not applied?"

packages/cli/src/settings/SettingsStoreImpl.test.ts
```diff
@@ -0,0 +1,146 @@
...
+    await chmod(root, 0o500);
+    const result = await store.patch({ ui: { theme: 'nord' } });
+    await chmod(root, 0o700);
+    assert(result.isErr());
+    expect(result.error.kind).toBe('filesystem');
+    expect(readdirSync(root)).toEqual([]);
+  });
+});
+
+function createStore(): SettingsStoreImpl {
```
jcarver989: Per other comment, could return the async disposable symbol form this here and the tmpdir could get cleaned up automatically on drop.

packages/cli/src/settings/SettingsStoreImpl.test.ts
```diff
@@ -0,0 +1,146 @@
...
+  temporaryDirectories.push(root);
+  return new SettingsStoreImpl({
+    env: environment.build({ XDG_CONFIG_HOME: root }),
+    logger: pino({ level: 'silent' }),
+  });
+}
+
+async function writeSettings(store: SettingsStoreImpl, contents: string): Promise<void> {
+  await mkdir(join(store.path, '..'), { recursive: true });
+  await writeFile(store.path, contents);
```
jcarver989: Seems sort of odd that we'd have a  `writeSettings` method here because the non-test path also has functionality to write the settings file, so seems like the tests should use that instead.

packages/cli/src/settings/SettingsStoreImpl.ts
```diff
@@ -0,0 +1,126 @@
...
+import type { Environment } from '#src/environment.ts';
+
+const SETTINGS_FILE = 'settings.json';
+
+interface SettingsStoreImplOptions {
+  readonly env: Pick<Environment, 'HOME' | 'XDG_CONFIG_HOME'>;
+  readonly logger: Logger;
+}
+
+export class SettingsStoreImpl implements SettingsStore {
```
jcarver989: Rather than `xxxImpl` I think a nicer naming scheme would be something like `SettingsFileStore`, `InMemorySettingsStore`, `S3SettingsStore` ... etc (not implying we'd have these other stores, just using them as illustrative examples). 

One other minor comment here is that I'd expect the only thing that would be different across `SettingsStore` implementations is the storage layer. So it'd be slightly nicer to make `SettingsStore` a class that accepts a storage implementation with methods like `save()`, `load()`. That would allow you to share all the other logic (e.g. `applyPatch`) across all implementations.

jcarver989 (approved):

PR #259, opened 2026-08-11
feat(settings): pull the theme picker into a settings dialog

### LINKED ISSUE #251: feat: user settings file with versioned schema and per-harness namespaces

blimmer:
## Summary

Pulls the header's theme picker into a modal **settings dialog**, completing the settings work — closes #251. The dialog follows familiar settings-panel patterns: a left sidebar of sections (theme today; harness-specific settings like Claude Code's permission mode land as new entries), a content pane built from the reusable `SettingsSection` chrome, and explicit **Save / Cancel** instead of the persist-on-select wiring #256 shipped. Theme selections preview live and revert on cancel; dismissing with unsaved changes (Esc, overlay click, X) asks for confirmation before discarding. Adds the standard shadcn `dialog` primitive to `@contextbridge/ui`.

## Review focus

- **Dirty-dismiss interception**: the dialog's `open` is controlled, and `onOpenChange(false)` with unsaved changes opens a nested `AlertDialog` instead of closing. Radix's layer stack makes Esc dismiss only the top layer, so Esc → warning → Keep Editing → Esc → Discard works — covered by tests, but the layering approach is worth a look.
- **Failed-persist semantics**: after Save, `savedTheme` advances even if the server write fails — the existing "changes apply to this session only" banner is the signal, and the dialog is not re-dirtied. Deliberate (the previewed theme *is* applied for the session), but it could have gone the other way.
- `DialogContent` needs `max-w-none sm:max-w-none` because tailwind-merge can't strip upstream's `sm:max-w-lg` with a base-variant override alone.

## Commits

- [`a374995`](https://github.com/contextbridge/planbridge/commit/a374995b3041d1b73be4db34cf927ee33d50fc63) — restructure `ThemePicker` into `SettingsDialog` (sidebar nav + `SettingsSection` panes + `ThemeSection`/`ThemeOption`), add the `dialog` primitive to `@contextbridge/ui`, rewire `App` for preview/save/cancel semantics, rewrite the settings tests, and add `SettingsDialog` Storybook stories (`Default`, `Opened`, `UnsavedChangesWarning`).


packages/annotation/src/SettingsDialog.tsx
```diff
@@ -0,0 +1,186 @@
...
+  readonly savedThemePreference: ThemePreference;
+  readonly onPreviewTheme: (preference: ThemePreference) => void;
+  readonly onSave: (preference: ThemePreference) => void;
+}
+
+export function SettingsDialog({ savedThemePreference, onPreviewTheme, onSave }: SettingsDialogProps) {
+  const [open, setOpen] = useState(false);
+  const [discardDialogOpen, setDiscardDialogOpen] = useState(false);
+  const [activeSectionId, setActiveSectionId] = useState<SettingsSectionId>('theme');
+  const [draftTheme, setDraftTheme] = useState(savedThemePreference);
```
blimmer, line 58: Theme is kind of a one-off since it affects the system state here. Other settings (like "which mode do we go into upon plan approval") will be more set-and-forget until it's needed.

jcarver989 (approved):

PR #267, opened 2026-08-11
feat!: allow using different Claude Code permission modes (e.g., auto) when accepting a plan

### LINKED ISSUE #247: feat: add option to exit plan mode into auto mode in Claude Code

blimmer:
## Summary

Approving a plan always exited Claude Code into `acceptEdits`. The mode is now a persisted setting (`harnesses.claude.planApprovalMode`) with a picker under Settings → Claude Code in the review UI; the default is `auto`, which is also Claude Code's default. Closes #247. Supersedes #249 (the first commit keeps @jpalharini's authorship).

Breaking change: approvals now land in `auto` instead of `acceptEdits`, and once the new key is written, the settings file stops parsing on older `contextbridge` binaries.

## Review focus

The hook reads the mode from the settings store after the review returns rather than from the submission payload, so `AnnotationSubmission` stays kind-agnostic at the cost of a second read with its own abort path. Also check the sparse-file tests: a theme-only save must never write `planApprovalMode` into the file.

## Commits

- [`1d71284`](https://github.com/contextbridge/planbridge/commit/1d71284c75d07aa32a669fd83030fb82d3dd83b7) — refactor(cli): build formatter test submissions with the shared factory (from #249, authored by @jpalharini)
- [`5bb7e5e`](https://github.com/contextbridge/planbridge/commit/5bb7e5ec71dbd010bda65909acc735537d2e3ebb) — feat(shared): add harnesses.claude.planApprovalMode setting
- [`dbe1ab6`](https://github.com/contextbridge/planbridge/commit/dbe1ab6eaf4d9513348fa7afd83609b49a12ddc6) — feat(cli)!: exit approved plans into the configured permission mode
- [`273c447`](https://github.com/contextbridge/planbridge/commit/273c4477e2f4dc5985f54a9d91c229a5370f8c75) — refactor(annotation): move settings components into settings/
- [`8fd07e8`](https://github.com/contextbridge/planbridge/commit/8fd07e83441ad5e4928f77fb397a57cca5ae0b37) — feat(annotation): add Claude Code section to the settings dialog
- [`70d2bb1`](https://github.com/contextbridge/planbridge/commit/70d2bb1920f1160070704b0907be69588943a24d) — docs(website): document the plan approval mode setting


packages/shared/src/claudeSettingsSchema.ts
```diff
@@ -0,0 +1,9 @@
+import { z } from 'zod';
+
+// A subset of the Claude Agent SDK's PermissionMode union, minus 'plan'. This package must
+// not depend on the SDK; the subset stays honest through contextual typing where the mode
+// flows into a PermissionUpdate in packages/cli/src/formatters/plan/claudeHookResponse.ts.
+export const CLAUDE_PLAN_APPROVAL_MODES = ['auto', 'acceptEdits', 'default', 'dontAsk', 'bypassPermissions'] as const;
```
jcarver989: I thought we said that `default` wasn't an option we could set from the hook side of things? Perhaps `default` here means something else?

jcarver989 (approved):
Thanks, you can ignore my comment on `default` I saw in a later commit that it refers to a mode in CC that's actually named 'default', which is different than the default mode the user can configure in Claude's settings...

packages/shared/src/claudeSettingsSchema.ts
blimmer: It's a bad name - that's the "manual" mode which, in fact, is not a default anymore haha.

I've updated the mode to "manual" in the UI that users see. I also removed `dontAsk` and `bypassPermissions`, as they were not relevant.

PR #271, opened 2026-08-19
test: use shared createDeferred helper in harnessDiscovery test

ghost:
## What & why

The **testing-patterns** convention (`.agents/memories/testing-patterns.md`, mirrored in the root `AGENTS.md`) states:

> **Use the shared deferred-promise helper; do not hand-roll it.** Tests that need manual promise resolution should import `createDeferred` from `@contextbridge/shared/testHelpers`. Do not recreate local `Deferred` types or `new Promise` wrappers in individual test files.

`packages/instrumentation/src/node/harnessDiscovery.test.ts` violated this: it declared its own module-local `createDeferred()` (a hand-rolled `new Promise` + captured-resolver wrapper) instead of importing the shared helper.

This change:
- Imports `createDeferred` from `@contextbridge/shared/testHelpers`.
- Updates the call site to `createDeferred<void>()` (the shared helper is generic).
- Deletes the duplicated local `createDeferred` function.

Behavior is unchanged — the shared helper returns the same `{ promise, resolve }` shape (plus an unused `reject`), and the test used only `.promise` and `.resolve()`.

## Verification

Run against the real bun binary (the asdf `bun` shim is broken in this environment):
- `bun test src/node/harnessDiscovery.test.ts` — 5 pass / 0 fail, including the `defers detection and capture` test that exercises the deferred.
- `tsc --noEmit` (package typecheck) — clean.
- `eslint --max-warnings 0` on the file — clean.
- `prettier --check` on the file — clean.

## Scope / notes for reviewer

This was the single most clearly-documented, unambiguous, safely-fixable convention violation. I fixed **all** instances of this class: this test file was the only test that hand-rolled a deferred. Note that `packages/cli/src/annotation/runAnnotation.ts`, `packages/cli/src/commands/review.ts`, and `packages/server/src/annotation.ts` also construct `new Promise` with captured resolvers, but those are **production** signal-handler / server-orchestration promises (they need `reject` for SIGINT handling), not test code — the testing convention explicitly scopes to tests, so they were intentionally left alone.

### Other violation classes noticed (deferred to future runs)

The repo adheres to its conventions very well; remaining candidates are judgment-heavy and not safely mechanizable in a single conservative pass:

- **neverthrow: prefer combinators over manual `isErr()` / try/catch** — several `if (result.isErr()) return err(...)` chains and imperative `try/catch` blocks exist (e.g. `extractAssets.ts`, `commands/*.ts`, `server/routes/*`). The rule explicitly permits `isErr()` and try/catch "as the exception" and for genuinely imperative code, so blanket refactoring risks behavior changes.
- **testing: prefer `toMatchObject` over field-by-field `expect().toBe()` runs** — many test files have long runs of per-field assertions, but the rule keeps exceptions for orthogonal behavior / specialized matchers, making it non-mechanical.
- **bun-native APIs** — a few `readFileSync`/`writeFileSync` calls in bun-only packages, but most are wrapped in `fromThrowable` tight-sync helpers the rule explicitly allows.
- **ctx reach-through** — `getInstallerRefreshScopes` in `commands/update.ts` reaches through `ctx.logger` once while its sibling destructures `const { ... } = ctx;`; a minor single-use inconsistency.

### DIFF
--- a/packages/instrumentation/src/node/harnessDiscovery.test.ts
+++ b/packages/instrumentation/src/node/harnessDiscovery.test.ts
@@ -4,6 +4,7 @@ import { dirname, join } from 'node:path';
 import type { BaseContext } from '@contextbridge/context';
 import { buildInfo as buildInfoFactory } from '@contextbridge/context/testFactories';
 import { fakeBaseContext } from '@contextbridge/context/testHelpers';
+import { createDeferred } from '@contextbridge/shared/testHelpers';
 import { type Instant, Temporal } from '@contextbridge/shared/time';
 import { afterEach, beforeEach, describe, expect, it } from 'bun:test';
 import { type FakeAnalytics, createFakeAnalytics } from '@contextbridge/instrumentation/testHelpers';
@@ -111,7 +112,7 @@ describe('reportHarnessDiscovery', () => {
   });
 
   it('defers detection and capture so CLI startup is not blocked', async () => {
-    const deferred = createDeferred();
+    const deferred = createDeferred<void>();
     let getHarnessDetectionsCalled = false;
     const { ctx, analytics, options } = createReportFixture(root, {
       clock: () => now,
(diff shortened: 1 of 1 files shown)


blimmer:
ContextBot has [been retired](https://contextbridge.ai/contextbot/). Closing.

